Issue #12, opened 2021-05-25
Improve Documentation


nniehoff:
### Environment

- Python version: 3.6
- Nautobot version: 1.0.1
- nautobot-circuit-maintenance version: 0.1.1

In the add notifications form I am prompted for a Url with no description and source type is either gmail or unknown.  A minimal description of the Url field and some documentation around configuring other email sources would be very beneficial.

### Expected Behavior

The documentation should provide examples for configuring a POP3 or an IMAP mail server at a minimum.

### Observed Behavior

Documentation refers to URL but nothing further

chadell:
fixed in #13

PR #13, opened 2021-05-27
Improve Notification Source model and documentation

chadell:
From issues #12 and #15 , we have to prepare the Notification Source to accommodate other integrations.
In this PR:
* Notification Sources configuration is moved from the UI to the `configurations.py`, so Nautobot is not storing secrets and eventually could fetch them externally.
* `NotificationSource` class is simplified to contain only the `alias` and the `providers`, and it's autopopulated from the config file in the `PluginConfig.ready`
* Moved from the **email** class to **source** so we could accommodate more custom types in the future extending the class.
* Fixed issue in `NotificationSource` view template


nautobot_circuit_maintenance/handle_notifications/email_helper.py
```diff
@@ -39,17 +39,11 @@ class Config:
     def __post_init__(self):
         """Post init validation."""
         if self.user is None:
```
glennmatthews, line 41: I'm surprised that these checks are needed, given that `user`, `password`, and `imap_url` are explicitly declared above as `str` (and not `Optional[str]`). Doesn't Pydantic automatically reject initialization with None values for any of these parameters?

nautobot_circuit_maintenance/models.py
```diff
@@ -241,27 +241,33 @@ class NotificationSource(OrganizationalModel):
 
     # Mark field as private so that it doesn't get included in ChangeLogging records!
     _password = encrypt(models.CharField(max_length=100))
-    source_id = models.EmailField(max_length=100, unique=True, help_text="Email address used as identifier.")
-    url = models.CharField(max_length=200)
+    source_id = models.EmailField(
+        max_length=100, unique=True, help_text="Identifier (i.e. email address) to use to authenticate."
```
glennmatthews, line 245: Possibly add `verbose_name="Source ID"` so that the form presents it as such instead of as `Source id`?

nautobot_circuit_maintenance/models.py
```diff
@@ -241,27 +241,33 @@ class NotificationSource(OrganizationalModel):
...
+    )
+    url = models.CharField(
+        max_length=200,
+        help_text="URL to reach the Notification Source.",
```
glennmatthews, line 249: Possibly add `verbose_name="URL"` so that the form doesn't render it as `Url`?

nautobot_circuit_maintenance/handle_notifications/email_helper.py
chadell: I agree with you... the validation was slightly different in the past... and the inercia left them here

glennmatthews:
Not sure I agree with the change from URL (which includes scheme, address, port, and path) to simply server + port - this seems like a loss of functionality to me. What's the benefit of setting `("gmail.com", 993)` versus being able to set something like `"imap://gmail.com:993"`?

chadell:
> Not sure I agree with the change from URL (which includes scheme, address, port, and path) to simply server + port - this seems like a loss of functionality to me. What's the benefit of setting `("gmail.com", 993)` versus being able to set something like `"imap://gmail.com:993"`?

You point is valid. I proposed the change to be able to specify the port, but as you say we can do it within the URL. So I will go back to the URL, but make it more capable as the implementation was only taking into account the address part, not even the port (using IMAP default), and the idea of this improvement is to open the door to connect to an API to get notifications from other sources than IMAP and the port should be configurable.

nniehoff:
> Not sure I agree with the change from URL (which includes scheme, address, port, and path) to simply server + port - this seems like a loss of functionality to me. What's the benefit of setting `("gmail.com", 993)` versus being able to set something like `"imap://gmail.com:993"`?

I guess I look at URL as being odd... when configuring IMAP servers URL might be standard in some libraries but from a user standpoint when configuring IMAP, think Outlook, I would expect server, port, protocol, SSL etc being options.  If we want to allow URL we need to document the heck out of it and provide a link from the form to the doc as most users are going to be wondering how to format the URL.

chadell:
I improved the README and also the form help:
image

I also believe that adding everything in the URL makes it easier to integrate with an API endpoint in the future.

nautobot_circuit_maintenance/models.py
```diff
@@ -240,37 +239,42 @@ class NotificationSource(OrganizationalModel):
...
-    url = models.CharField(max_length=200)
-    source_type = models.CharField(
-        default=NotificationSourceServerChoices.UNKNOWN,
-        max_length=50,
-        choices=NotificationSourceServerChoices,
-        null=True,
-        blank=True,
-        help_text="Type of source integration to retrieve the notifications.",
+    _password = encrypt(models.CharField(max_length=100, verbose_name="Password"))
+    source_id = models.EmailField(
```
glennmatthews, line 243: Come to think of it, this is a bit of a weird name. Why not something like "username" or "account"?

nautobot_circuit_maintenance/models.py
```diff
@@ -240,37 +239,42 @@ class NotificationSource(OrganizationalModel):
...
+        max_length=100,
```
glennmatthews, line 244: Possibly increase this to 255 as long as you're making changes here - email addresses can definitely be longer than 100 characters.

nautobot_circuit_maintenance/models.py
```diff
@@ -240,37 +239,42 @@ class NotificationSource(OrganizationalModel):
...
+        unique=True,
+        verbose_name="Source ID",
+        help_text="Identifier (i.e. email address) to use to authenticate.",
+    )
+    url = CustomURLField(
+        max_length=200,
```
glennmatthews, line 250: Might also need to increase this limit substantially once non-IMAP URLs are supported, but probably fine for the moment.

chadell:
@nniehoff @glennmatthews I would appreciate your feedback as I plan to do next release on 9th of June, before publishing the Blog about this project. Thanks in advance

nautobot_circuit_maintenance/migrations/0002_imporove_notification_source.py
```diff
@@ -0,0 +1,46 @@
+# Generated by Django 3.1.10 on 2021-05-31 04:26
```
glennmatthews, line 1: nitpick: `0002_improve_notification_source.py` (not `imporove`)

glennmatthews (approved):

chadell:
@glennmatthews I tried to capture last changes in the PR comment, but the main change are:
* `NotificiationSource` is now much simpler on the DB, only an `alias`+`providers`, and it's auto-populated from the configs file (so they can't be created or deleted, only edited to modify the providers)
* Removed old references to `settings` (initial name for `NotificationSource`) or email
* Created a new `Source` class that will be used to implement external integrations with custom attributes and with a public method `receive_notifications` that is used to retrieve emails or other notifications sources in the future.

So, in short, we are removing the secrets from the DB and passing as config via ENV, following the Nautobot idea to extract secrets, and be ready to integrate with external secret services in the future.

README.md
```diff
@@ -20,51 +20,84 @@ To ensure Circuit Maintenance is automatically re-installed during future upgrad
...
-- Providers: list of Ciruit Providers that will be tracked.
+In the Circuit Maintenance plugin UI section, there is a **Notification Sources** button (yellow) where you can configure the Notification Sources to track new circuit maintenance notifications from specific providers.
 
-> [How to setup Gmail with App Passwords](https://support.google.com/accounts/answer/185833)
+Because the Notification Sources are defined by the configuration, you can only view and edit `providers`, but not `add` or `delete` new Notification Sources via UI or API.
 
 ### 3. Run Handle Notifications Job
 
-There is an asynchronous task defined as a **Nautobot Job**, **Handle Circuit Mainentance Notifications** that will connect to the emails sources defined under the Settings section (step above), and will fetch new notifications received since the last notification was fetched.
+There is an asynchronous task defined as a **Nautobot Job**, **Handle Circuit Mainentance Notifications** that will connect to the emails sources defined under the Notification Sources section (step above), and will fetch new notifications received since the last notification was fetched.
```
glennmatthews, line 100:
```suggestion
There is an asynchronous task defined as a **Nautobot Job**, **Handle Circuit Maintenance Notifications** that will connect to the emails sources defined under the Notification Sources section (step above), and will fetch new notifications received since the last notification was fetched.
```

nautobot_circuit_maintenance/handle_notifications/sources.py
```diff
@@ -209,26 +229,32 @@ def get_notifications_from_email(
...
+                        message=f"No Notification Source configuration available or supported scheme for {notification_source.alias}.",
+                    )
+                    continue
+            except ValidationError as validation_error:
+                logger.log_warning(
+                    message=f"Notification Source {notification_source.alias} is not matching class expectations: {validation_error}",
+                )
+                continue
+
+            rawnotification = imap_conn.receive_notifications(logger, restrict_emails, since)
```
glennmatthews, line 248:
```suggestion
                source = Source.init(alias=notification_source.alias)
                if not source:
                    logger.log_warning(
                        message=f"No Notification Source configuration available or supported scheme for {notification_source.alias}.",
                    )
                    continue
            except ValidationError as validation_error:
                logger.log_warning(
                    message=f"Notification Source {notification_source.alias} is not matching class expectations: {validation_error}",
                )
                continue

            rawnotification = source.receive_notifications(logger, restrict_emails, since)
```

Also seems like a bit incorrect to have `restrict_emails` as a parameter for generic Sources.

nautobot_circuit_maintenance/tests/test_sources.py
```diff
@@ -0,0 +1,177 @@
...
+        self.assertEqual(source_instance.url, SOURCE_1["url"])
+        self.assertEqual(source_instance.user, SOURCE_1["account"])
+        self.assertEqual(source_instance.password, SOURCE_1["secret"])
+        self.assertEqual(source_instance.imap_server, "example.com")
+        self.assertEqual(source_instance.imap_port, 993)
+
+    def test_source_factory_nonexistent_alias(self):
+        """Validate Factory pattern for non existent alias."""
+        source_instance = Source.init(alias="non existent alias")
+        self.assertEqual(source_instance, None)
```
glennmatthews, line 52: Would it be better for `init` to raise an exception instead of returning None in these cases?

glennmatthews (commented):
I'm not entirely sure I follow the full scope of these changes but I think they look like the right direction in general.

nautobot_circuit_maintenance/handle_notifications/sources.py
chadell: yes... restrict_emails or `senders` from the signature was just an optimization to not fetch ALL the emails from an inbox (the same as `since` that limits the time scope). For Gmail API I guess we could reuse both as the API will provide a way to filter emails IIRC, but for more generic notification sources, when the time comes, this data should be ignored and another way to map to providers should be implemented.
I will add a TODO note to take care of this eventually.

nautobot_circuit_maintenance/tests/test_sources.py
chadell: Yes, I will implement it with an exception: https://www.kim-lindhard.com/is-it-ok-for-a-factory-to-return-null/

nautobot_circuit_maintenance/models.py
```diff
@@ -239,38 +235,30 @@ def to_csv(self):
...
-    source_id = models.EmailField(max_length=100, unique=True, help_text="Email address used as identifier.")
-    url = models.CharField(max_length=200)
-    source_type = models.CharField(
-        default=NotificationSourceServerChoices.UNKNOWN,
-        max_length=50,
-        choices=NotificationSourceServerChoices,
-        null=True,
-        blank=True,
-        help_text="Type of source integration to retrieve the notifications.",
+    alias = models.CharField(
```
glennmatthews, line 238: I wonder if instead of `alias` we should use `name` (and/or `slug`?) to be more consistent with the rest of Nautobot? I don't feel strongly about this, just a thought.

nautobot_circuit_maintenance/__init__.py
```diff
@@ -23,6 +24,22 @@ def custom_field_extension(sender, **kwargs):  # pylint: disable=unused-argument
...
+    For now, we create the aliases in the DB but the secrets are fetched via ENV.
+    """
+    # pylint: disable=import-outside-toplevel
+
+    from nautobot_circuit_maintenance.models import NotificationSource
+
+    for notification_source in settings.PLUGINS_CONFIG.get("nautobot_circuit_maintenance", {}).get(
+        "notification_sources", []
+    ):
+        NotificationSource.objects.get_or_create(alias=notification_source["alias"])
```
glennmatthews, line 40: This logic probably should be refined so that when an entry is removed from `notification_sources` it gets deleted from Nautobot as well?

nautobot_circuit_maintenance/urls.py
```diff
@@ -100,20 +100,12 @@
         name="parsednotification",
     ),
     # Notification Source
-    path("settings/", views.NotificationSourceListView.as_view(), name="notificationsource_list"),
-    path("settings/add/", views.NotificationSourceEditView.as_view(), name="notificationsource_add"),
-    path("settings/<uuid:pk>/edit/", views.NotificationSourceEditView.as_view(), name="notificationsource_edit"),
-    path("settings/<uuid:pk>/delete/", views.NotificationSourceDeleteView.as_view(), name="notificationsource_delete"),
+    path("source/", views.NotificationSourceListView.as_view(), name="notificationsource_list"),
+    path("source/<uuid:pk>/edit/", views.NotificationSourceEditView.as_view(), name="notificationsource_edit"),
+    path("source/<uuid:pk>/", views.NotificationSourceView.as_view(), name="notificationsource"),
```
glennmatthews, line 105: Now that we have a `slug` for `NotificationSource`, could use it rather than the `pk` in these URL patterns.

nautobot_circuit_maintenance/models.py
```diff
@@ -239,38 +235,31 @@ def to_csv(self):
...
+    slug = models.SlugField(max_length=100, unique=True)
     providers = models.ManyToManyField(
         Provider,
-        help_text="The Provider(s) to which this notification source applies.",
+        help_text="The Provider(s) that this Notification Source applies to.",
         blank=True,
     )
 
-    csv_headers = ["source_id", "_password", "url"]
+    csv_headers = ["name", "slug"]
```
glennmatthews, line 250: Based on `to_csv` below, I'd think this should be:
```suggestion
    csv_headers = ["name", "slug", "providers"]
```

glennmatthews (approved):

PR #23, opened 2021-06-10
Abstract Notification processing and RawNotification model

chadell:
The goal of the PR was to start abstracting how do we handle Notifications from external sources, with these changes:
* Simplify the `get_notifications` function extracting the `Providers` validation to each `Source` class as this will depend on each type of integration, and fix a logic issue.
* `RawNotification.souce` is now a foreingkey to a `NotificationSource` and the `sender` is optional


nautobot_circuit_maintenance/handle_notifications/sources.py
```diff
@@ -72,7 +94,9 @@ class IMAP(Source):
     imap_server: str
     imap_port: int = 993
 
-    session: Union[imaplib.IMAP4_SSL, None] = None
+    _session: Union[imaplib.IMAP4_SSL, None] = None
```
glennmatthews, line 97: What's the reason for changing this to a private variable?

nautobot_circuit_maintenance/handle_notifications/sources.py
```diff
@@ -193,9 +215,53 @@ def receive_notifications(
         self.close_session()
         return received_notifications
 
+    def validate_providers(self, logger: Job, notification_source: NotificationSource, since_txt: str) -> bool:
+        """Method to validate that the NotificartionSource has attached Providers.
```
glennmatthews, line 219:
```suggestion
        """Method to validate that the NotificationSource has attached Providers.
```

nautobot_circuit_maintenance/handle_notifications/sources.py
```diff
@@ -193,9 +215,53 @@ def receive_notifications(
...
+            )
+            return False
+
+        for provider in notification_source.providers.all():
+            for custom_field, value in provider.get_custom_fields().items():
+                if custom_field.name == "emails_circuit_maintenances" and value:
+                    self._emails_to_fetch.extend(value.split(","))
+                    providers_with_email.append(provider.name)
+                else:
+                    providers_without_email.append(provider.name)
```
glennmatthews, line 243: This logic seems incorrect as it will add a provider that has multiple custom fields to both `providers_with_email` and `providers_without_email`.

```suggestion
            custom_fields = provider.get_custom_fields()
            emails = custom_fields.get("emails_circuit_maintenance")
            if emails:
                self._emails_to_fetch.extend(emails.split(","))
                providers_with_email.append(provider.name)
            else:
                providers_without_email.append(provider.name)
```

nautobot_circuit_maintenance/handle_notifications/sources.py
```diff
@@ -193,9 +215,53 @@ def receive_notifications(
...
+
+        if providers_without_email:
+            logger.log_warning(
```
glennmatthews, line 246: Possibly change this one to `log_info`? I suspect it'll be pretty normal for us to have some providers without email.

nautobot_circuit_maintenance/handle_notifications/sources.py
```diff
@@ -193,9 +215,53 @@ def receive_notifications(
...
+                message=f"Skipping {', '.join(providers_without_email)} because these providers has no email configured.",
```
glennmatthews, line 247:
```suggestion
                message=f"Skipping {', '.join(providers_without_email)} because these providers have no email configured.",
```

nautobot_circuit_maintenance/handle_notifications/sources.py
```diff
@@ -193,9 +215,53 @@ def receive_notifications(
...
+            )
+
+        if not providers_with_email:
+            logger.log_warning(
+                message=f"Skipping this email account because none of the providers ({', '.join(providers_with_email)}) have at least one email defined.",
```
glennmatthews, line 252: `providers_with_email` is empty in this case so it doesn't need to be embedded in the log message
```suggestion
                message=f"Skipping this email account because none of the providers have at least one email defined.",
```

nautobot_circuit_maintenance/handle_notifications/sources.py
```diff
@@ -251,13 +284,15 @@ def get_notifications(  # pylint: disable=too-many-locals, too-many-branches
                 logger.log_warning(message=value_error)
                 continue
 
-            rawnotification = imap_conn.receive_notifications(logger, restrict_emails, since)
-            received_notifications.extend(rawnotification)
+            if source.validate_providers(logger, notification_source, since_txt):
+                rawnotification = source.receive_notifications(logger, since)
+                received_notifications.extend(rawnotification)
```
glennmatthews, line 289: `rawnotification` is a list, not a single item, correct?
```suggestion
                raw_notifications = source.receive_notifications(logger, since)
                received_notifications.extend(raw_notifications)
```

nautobot_circuit_maintenance/handle_notifications/sources.py
chadell: I defined as warning because if those providers are "attached" to this NotificationSource, it doesn't make any sense because without a `source_email` none of the emails fetched by this IMAP source will be related to this provider, so it would like to signal it as warning to catch the attention from the operator.
Note: the providers taken into account are the ones related to this NotificationSource, that has been instantiated as IMAP. Other source types, for instance a rest API, would not use the same `validate_providers` logic, will have a custom one that won't check for  `source_emails`

nautobot_circuit_maintenance/handle_notifications/sources.py
chadell: yes, I agree with you

glennmatthews (approved):

nautobot_circuit_maintenance/handle_notifications/sources.py
```diff
@@ -193,9 +215,65 @@ def receive_notifications(
...
+                message="Skipping this email account no providers were defined.",
+            )
+            return False
+
+        for provider in notification_source.providers.all():
+            provider_emails = ""
+            for custom_field, value in provider.get_custom_fields().items():
+                if custom_field.name == "emails_circuit_maintenances" and value:
+                    provider_emails = value
+                    break
```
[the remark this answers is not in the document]
chadell: `provider.get_custom_fields()` returns and `OrderedDict` that doesn't work fine with `.get("emails_circuit_maintenance", "")`... but maybe it's me who is missing something 🤒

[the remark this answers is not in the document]
glennmatthews: Really? `OrderedDict` is a subclass of `dict`, it ought to support the same syntax.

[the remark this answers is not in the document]
chadell: it actually works as you say and I expected, but the code was not behaving well... will double check to validate it
```py
>>> from collections import OrderedDict
>>> numbers = OrderedDict()
>>> numbers["one"] = 1
>>> numbers.get("one")
```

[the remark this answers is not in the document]
chadell: @glennmatthews -> b1392c1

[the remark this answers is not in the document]
glennmatthews: Ah, okay, I hadn't remembered that `get_custom_fields()` returns a dict of `CustomField` objects. That makes sense then. The way you've coded it is fine; the alternative would be to use `provider.cf` (which is an ordinary key-value dictionary) instead of `provider.get_custom_fields()`, but either way is fine.

glennmatthews (approved):

Issue #15, opened 2021-06-05
Support Gmail API Notification Source


chadell:
### Proposed Functionality

Add Gmail API as a supported Notification Source, supporting OAuth authentication.

See: https://www.geeksforgeeks.org/how-to-read-emails-from-gmail-using-gmail-api-in-python/

Because of the OAuth nature, the "Application", the Plugin, will require:
* the first time we set it up a manual login verification, later we could keep refreshing the token. But, we should offer a way to the user to "activate" this source. Maybe, a "TEST/VALIDATE" option could serve this goal and also provide an easy way to validate the credentials for other methods, such as IMAP.
* the way that we identify to the API is more complex than just an account id (i.e. email) and password, so it contains a "client_id", and "project_id", and a "client_secret". This means that the model should be modified to support these more complex cases.
* it should work across service restarts

### Use Case

Offer an alternative to IMAP Notification Source

chadell:
Fixed in #26

PR #25, opened 2021-06-11
Add authentication_status attribute to NotificationSource

chadell:
Add a validate authentication for notification sources to give a quick understanding of it works or not before actually fetching notifications.
My goal behind this PR is to facilitate adding new `NotificationSources` as pointed out in issue #15  which I will start after this PR.

image


glennmatthews (commented):
General comment/question - does it actually make sense to store authentication status as a DB field? Or is it something that should be dynamically generated on request (like a device's LLDP neighbors) since it isn't really something that persists -- authentication may succeed **at this moment** but silently change to failing later because the account's credentials were changed or such?

chadell:
> General comment/question - does it actually make sense to store authentication status as a DB field? Or is it something that should be dynamically generated on request (like a device's LLDP neighbors) since it isn't really something that persists -- authentication may succeed **at this moment** but silently change to failing later because the account's credentials were changed or such?

This is actually a good point... I implemented this because I was foresee that to use Oauth later I will need an initial login from the user to generate the first token and keep refreshing... so I would force the operator to login it at least once.
But this won't actually work between application resets... so I believe that I will have to use Google `Service Accounts`...
and this rational is to conclude that you are right, we can't assume that what was authenticated in the past is it still  in the present.

So I will keep the feature but without actually storing it, just as a dynamic data

nautobot_circuit_maintenance/handle_notifications/sources.py
```diff
@@ -117,6 +127,27 @@ def close_session(self):
...
+        """
+        try:
+            self.open_session()
+            self.close_session()
+            is_authenticated = True
+            message = "Test OK"
+        except Exception as exc:
+            is_authenticated = False
+            if isinstance(exc.args[0], bytes):
+                message = str(exc.args[0].decode())
```
glennmatthews, line 146: What's the scenario that this `if` condition is handling? I would have thought `str(exc)` would generally be the desired output?

glennmatthews (approved):
Looks good!

nautobot_circuit_maintenance/handle_notifications/sources.py
chadell: I was getting a `str` from `bytes` such as `b'ABC'`, so `decoding` first removes the `b''` (maybe there is a better way :) )

PR #27, opened 2021-06-18
Small UX improvements

chadell:
# New line for Notification Sources and a button to redirect to the Update Circuit Maintenance Job
image

# Extended View for Notification Source, with account ID and type
image

### DIFF
--- a/nautobot_circuit_maintenance/views.py
+++ b/nautobot_circuit_maintenance/views.py
@@ -1,5 +1,5 @@
 """Views for Circuit Maintenance."""
-from django.shortcuts import get_object_or_404, render
+from django.shortcuts import get_object_or_404, render, redirect
 from nautobot.core.views import generic
 from nautobot.circuits.models import Provider
 from nautobot_circuit_maintenance import filters, forms, models, tables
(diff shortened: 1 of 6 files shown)


chadell:
@jedelman8 adding some of the UX improvements we talked about

nautobot_circuit_maintenance/views.py
```diff
@@ -70,6 +70,20 @@ class CircuitMaintenanceBulkDeleteView(generic.BulkDeleteView):
...
+
+    queryset = models.CircuitMaintenance.objects.all()
+
+    def get(self, request, *args, **kwargs):
+        """Custom GET to run a the Job."""
+        class_path = (
+            "plugins/nautobot_circuit_maintenance.handle_notifications.handler/HandleCircuitMaintenanceNotifications/"
+        )
+
+        return redirect(f"/extras/jobs/{class_path}")
```
glennmatthews, line 84: Probably better to use `redirect(reverse("extras:job", kwargs={"class_path": class_path}))` here

glennmatthews (dismissed):

nautobot_circuit_maintenance/views.py
chadell: nicer!

PR #28, opened 2021-06-21
Bump Circuit Maintenance Parser v1.2.1

chadell:
Main goal of this PR is to adopt the v1.2.0 of `circuit-maintenance-parser` that introduces multiple parsers per Provider and removes the class `MaintenanceNotification` and now it uses a `Provider` one but I made it a bit more independent from the library implementation, limiting it's usage within a method `process_raw_notification`

In order to adapt the code in the plugin, we created the `MaintenanceNotification` class as a container of all the data we will use in this plugin when fetching notifications from external Sources, depending on the multiple data types supported for each provider. So, for instance, Telstra supports ICalendar and HTML, so for each notification from Telstra we will look for payloads of both types, and try to parse them in order.

Extras:
* Fix the NotificationSource Validation view to represent all the data


nautobot_circuit_maintenance/handle_notifications/sources.py
```diff
@@ -287,26 +296,34 @@ def fetch_email(self, logger: Job, msg_id: bytes, since: Optional[int]) -> Optio
...
+                logger.log_debug(
+                    message=f"Payload type {provider_data_type} not found in email payload.",
+                )
+                continue
+            break
         else:
             logger.log_warning(
-                message=f"Payload type {provider_data_type} not found in email payload.",
+                message=f"Payload types {', '.join(provider_data_types)} not found in email payload.",
             )
```
glennmatthews, line 319: I'm actually not sure this is working as it's supposed to, do you want to only get the first payload that matches the first provider_data_type? Could definitely do with some comments about the intended flow, and consider again breaking this into a helper function to simplify the flow control logic.

chadell: good catch!

nautobot_circuit_maintenance/handle_notifications/sources.py
```diff
@@ -425,6 +425,16 @@ def fetch_email(self, logger: Job, msg_id: bytes, since: Optional[int]) -> Optio
...
         """
+
+        def get_raw_payload_from_parts(parts, provider_data_type):
+            """Helper function to extract the raw_payload from a multiple parts via a recursive call."""
+            for part_inner in parts:
+                for header_inner in part_inner["headers"]:
+                    if header_inner.get("name") == "Content-Type" and provider_data_type in header_inner.get("value"):
+                        return self.extract_raw_payload(part_inner["body"], msg_id)
+                    elif header_inner.get("name") == "Content-Type" and "multipart" in header_inner.get("value"):
+                        return get_raw_payload_from_parts(part_inner["parts"], provider_data_type)
```
glennmatthews, line 436: Probably would be good style-wise to have an explicit `return None` at the end of the function if none of the earlier conditions are met.

nautobot_circuit_maintenance/handle_notifications/sources.py
```diff
@@ -287,26 +296,34 @@ def fetch_email(self, logger: Job, msg_id: bytes, since: Optional[int]) -> Optio
...
+        for provider_data_type in provider_data_types:
+            for part in email_message.walk():
+                if part.get_content_type() == provider_data_type:
+                    raw_payloads.append(part.get_payload())
+                    break
+            else:
+                logger.log_debug(
+                    message=f"Payload type {provider_data_type} not found in email payload.",
+                )
+                continue
```
glennmatthews, line 314: Unnecessary `continue`, I think?

glennmatthews (approved):

PR #36, opened 2021-07-01
Gmail API Oauth2

chadell:
# PR changes

* Created a new class `GmailAPI` that includes all the Gmail API interactions (taken from old `GmailAPIServiceAccount` directly)
* For `GmailAPIServiceAccount` and new `GmailAPIOauth`, the only customisation is the `load_credentials`, because in the first case it's taken directly from the credentials_file, and  for the second, there is a OAuth2.0 workflow.
* The OAuth workflow has been implemented with two new urls/views, `google_authorize` (to initiate the workflow after raising a custom exception `RedirectAuthorize` that contians the proper redirect URL and source name) and `google_oauth2callback ` to receive the redirect from Google after successful authentication. 
* The way to keep track of the token received is by using a new DB binary field hidden in the `Notification Sources`, and using always the `consent` and `access_type="offline"` to enable renewal of the password after a first successful authorization.
* Adapted and extended test for the new classes.
* Removed unused file `development/.docker-compose.yml`

## TODO

- [x] Document Readme setup in Google
- [x] Move dynamic secrets into DB
- [x] Fix tests
- [x] Organize Exceptions
- [x] Add new tests
- [x] test app for several days


chadell:
as a not, I tried to "validate" authentication 7 days later and it worked without prompting for credentials!

nautobot_circuit_maintenance/handle_notifications/sources.py
```diff
@@ -128,12 +135,25 @@ def init(cls: Type[T], name: str) -> Type[T]:
...
-            )
+            creds_filename = config.get("credentials_file")
+            if not creds_filename:
+                raise ValueError(f"Credentials_file for {name} not found in PLUGINS_CONFIG.")
+
+            if not os.path.isfile(creds_filename):
+                raise ValueError(f"Credentials_file {creds_filename} for {name} is not available.")
+
+            with open(creds_filename) as crededentials_file:
+                credentials = json.load(crededentials_file)
```
glennmatthews, line 146:
```suggestion
            with open(creds_filename) as credentials_file:
                credentials = json.load(credentials_file)
```

nautobot_circuit_maintenance/handle_notifications/sources.py
```diff
@@ -128,12 +135,25 @@ def init(cls: Type[T], name: str) -> Type[T]:
...
+                if credentials.get("type") == "service_account":
+                    gmail_api_class = GmailAPIServiceAccount
+                else:
+                    gmail_api_class = GmailAPIOauth
```
glennmatthews, line 150: Perhaps this could be an `else if` to check specifically for OAuth2 type, and `else` raise an exception for an unknown/unsupported type?

nautobot_circuit_maintenance/handle_notifications/sources.py
```diff
@@ -516,6 +529,54 @@ def receive_notifications(self, logger: Job, since: int = None) -> Iterable[Main
...
+    See: https://developers.google.com/identity/protocols/oauth2/web-server
+    """
+
+    def load_credentials(self):
+        """Load Gmail API OAuth credentials."""
+        notification_source = NotificationSource.objects.get(name=self.name)
+        try:
+            self.credentials = notification_source.get_token()
+        except EOFError:
+            # _token has not been initialized yet
```
glennmatthews, line 554: Should we log a message here? Not sure if that should be a debug message or a warning message though?

nautobot_circuit_maintenance/models.py
```diff
@@ -1,9 +1,11 @@
 """Models for Circuit Maintenance."""
 import hashlib
+import pickle  # nosec
 from django.core.exceptions import ValidationError
 from django.db import models
 from django.urls import reverse
 from django.utils.timezone import now
+from django_cryptography.fields import encrypt
```
glennmatthews, line 8: We're having a lot of ongoing pain with the use of `django_cryptography` in Nautobot core. I'd highly recommend against adding any new uses of this library.

nautobot_circuit_maintenance/tests/test_sources.py
```diff
@@ -313,40 +328,70 @@ class TestGmailAPISource(TestCase):
...
-        self.source = NotificationSource.objects.get(name=SOURCE_GMAIL_API["name"])
+        NotificationSource.objects.exclude(name__in=[SOURCE_GMAIL_API_SERVICE_ACCOUNT["name"]]).delete()
+        self.source = NotificationSource.objects.get(name=SOURCE_GMAIL_API_SERVICE_ACCOUNT["name"])
 
-    def test_source_factory(self):
+        with open(SOURCE_GMAIL_API_SERVICE_ACCOUNT["credentials_file"], "w") as credentials_file:
+            json.dump({"type": "service_account"}, credentials_file)
+
+        with open(SOURCE_GMAIL_API_OAUTH["credentials_file"], "w") as credentials_file:
+            json.dump({}, credentials_file)
```
glennmatthews, line 344: Can we provide more "realistic" JSON data than these minimal dicts? What other keys are typically present in these files in reality?

nautobot_circuit_maintenance/views.py
```diff
@@ -287,6 +293,15 @@ def get(self, request, *args, **kwargs):
             message += f": {mess_auth}"
         except ValueError as exc:
             message = str(exc)
+        except RedirectAuthorize as exc:
+            return redirect(
+                reverse(
+                    f"plugins:nautobot_circuit_maintenance:{str(exc.url_name)}",
```
glennmatthews, line 299: What happens if `exc.url_name` doesn't refer to a valid URL pattern? Do we need to catch the `NoReverseMatch` exception and handle it gracefully?

nautobot_circuit_maintenance/views.py
```diff
@@ -299,3 +314,76 @@ def get(self, request, *args, **kwargs):
...
+    request.session["CLIENT_SECRETS_FILE"] = source.credentials_file
+    request.session["SCOPES"] = source.SCOPES
+    request.session["SOURCE_SLUG"] = slug
+
+    # Create flow instance to manage the OAuth 2.0 Authorization Grant Flow steps.
+    flow = google_auth_oauthlib.flow.Flow.from_client_secrets_file(
+        request.session["CLIENT_SECRETS_FILE"], scopes=request.session["SCOPES"]
+    )
+    # The URI created here must exactly match one of the authorized redirect URIs
+    # for the OAuth 2.0 client, which you configured in the API Console. If this
```
glennmatthews, line 333: Perhaps?

```suggestion
    # for the OAuth 2.0 client, which you configured in the Google Cloud Console. If this
```

nautobot_circuit_maintenance/views.py
```diff
@@ -299,3 +314,76 @@ def get(self, request, *args, **kwargs):
...
+        + reverse("plugins:nautobot_circuit_maintenance:google_oauth2callback")
+    )
+    authorization_url, state = flow.authorization_url(
+        # Enable offline access so that you can refresh an access token without
+        # re-prompting the user for permission. Recommended for web server apps.
+        access_type="offline",
+        # Enable incremental authorization. Recommended as a best practice.
+        include_granted_scopes="true",
+        # Asking for consent everytime is providing a refresh token that if not used only happens
+        # the first time.
```
glennmatthews, line 350: I'm having trouble parsing this sentence. Can you perhaps rephrase?

nautobot_circuit_maintenance/views.py
```diff
@@ -299,3 +314,76 @@ def get(self, request, *args, **kwargs):
...
+    return redirect(authorization_url)
+
+
+def google_oauth2callback(request):
+    """View to receive the callback from Google OAuth authorization flow."""
+    # Specify the state when creating the flow in the callback so that it can
+    # verified in the authorization server response.
+    state = request.session["state"]
+    flow = google_auth_oauthlib.flow.Flow.from_client_secrets_file(
+        request.session["CLIENT_SECRETS_FILE"], scopes=request.session["SCOPES"], state=state
```
glennmatthews, line 365: From a security standpoint, do we need to check that CLIENT_SECRETS_FILE is associated with a defined Source, or anything like that? I'm trying to understand whether this could potentially allow access to files that shouldn't be accessible or anything like that.

nautobot_circuit_maintenance/views.py
```diff
@@ -299,3 +314,76 @@ def get(self, request, *args, **kwargs):
...
+        + reverse("plugins:nautobot_circuit_maintenance:google_oauth2callback")
+    )
+    # Use the authorization server's response to fetch the OAuth 2.0 tokens.
+    authorization_response = request.build_absolute_uri()
+
+    flow.fetch_token(authorization_response=authorization_response)
+
+    # Store credentials in the session.
+    credentials = flow.credentials
+    notification_source = models.NotificationSource.objects.get(slug=request.session["SOURCE_SLUG"])
```
glennmatthews, line 381: Do we need to gracefully handle a possible `NotificationSource.DoesNotExist` exception?

nautobot_circuit_maintenance/models.py
chadell: which alternative I have to keep a dynamic secret shared across multiple Nautobot instances?

nautobot_circuit_maintenance/tests/test_sources.py
chadell: I could add more data, but I tried to keep it minimal to the actual code validation in factory method `init` (now updated with the oauth check). 
What would you like to test more regarding them?

nautobot_circuit_maintenance/views.py
chadell: Because the NotificationSource are defined by plugin configuration (user can't add or delete them), if we are at this point is because it is present, but adding a graceful handling will be good.

nautobot_circuit_maintenance/models.py
glennmatthews: IMHO, for the time being just store it in the DB unencrypted. You can make the UI and API refuse to retrieve/render the stored value to the user, if you'd like, as a first layer of defense, but anyone who has direct read access to the DB has enough privileges that I think this would be the least of your concerns security-wise.

In the longer term (i.e. https://github.com/nautobot/nautobot/issues/541) this would be something we'd probably want to reach out to an external system like Vault to store it in.

nautobot_circuit_maintenance/tests/test_sources.py
glennmatthews: Not sure. I guess I was assuming that the calls out to the Google libraries would need to do some sort of validation of this data, but perhaps not.

nautobot_circuit_maintenance/views.py
chadell: do you mean by an internally manipulated session? 
at the end, if the defined info in the file is not matching the info that Google is authorizing via the callback, it should fail.
Will think a bit more to see any potential problem

nautobot_circuit_maintenance/views.py
chadell: I don't see it as a possible with the current workflow, only by some unexpected error, but as commented before, adding graceful error handling is always good.

nautobot_circuit_maintenance/models.py
chadell: thanks for your feedback!

nautobot_circuit_maintenance/tests/test_sources.py
chadell: I just defined them to go over the factory pattern and get the proper class, I'm not doing any kind of integration tests with Google API at the moment.

nautobot_circuit_maintenance/views.py
glennmatthews: Since this is a View, any client (browser, Postman, etc.) could send a request to this endpoint - not just the Google server. Just want to make sure we don't have some sort of risk of spoofing here.

nautobot_circuit_maintenance/views.py
glennmatthews: As above - a browser or Postman could hit this view endpoint, not just Google servers, I think?

nautobot_circuit_maintenance/views.py
chadell: hopefully better now :)

nautobot_circuit_maintenance/views.py
chadell: I understand your concern, but the security problem is covered by the "state" info injected in the `google_authorize' view that it's verified in the authorization server response.
Obviously, as this is an open endpoint, what could happen is that if the session doesn't contain the expected keys: "state", "CLIENT_SECRETS_FILE", "SCOPES" and "SOURCE_SLUG", it will raise a `KeyError` exception, so I would handle this gracefully with a simple `get`

chadell:
@glennmatthews hopefully I've addressed your comments. Please, raise your hand if not.

nautobot_circuit_maintenance/models.py
```diff
@@ -190,6 +194,15 @@ def to_csv(self):
...
 
+    def set_token(self, value):
+        """Getter for _token."""
+        self._token = pickle.dumps(value)
+        self.save()
+
+    def get_token(self):
+        """Setter for _token."""
+        return pickle.loads(self._token)  # nosec
+
```
glennmatthews, line 205: Might be more Pythonic to use a `property` here, i.e.:

```python
@property
def token(self):
     return pickle.loads(self._token)

@token.setter
def token(self, value):
    self._token = pickle.dumps(value)
```

Also I wouldn't have the setter function automatically call `self.save()` as that doesn't match the general model API - would be better to explicitly call `save()` as needed.

README.md
```diff
@@ -121,6 +112,29 @@ PLUGINS_CONFIG = {
...
+3. Still under **APIs and Services**, in **Credentials**, create a new **Service Account** and save the credentials file generated.
+4. With Admin rights, edit the newly created Service Account and expand the **Show Domain-Wide Delegation** section. Enable Google Workspace Domain-wide Delegation and save the changes. Copy the Client ID shown.
+5. With Super Admin rights, open the [Google Workspace admin console](https://admin.google.com). Navigate to **Security**, **API controls**, and select the **Manage Domain Wide Delegation** at the bottom of the page.
+6. Add a new API client and paste in the Client ID copied earlier. In the **OAuth scopes** field add the scopes `https://www.googleapis.com/auth/gmail.readonly` and `https://mail.google.com/`. Save the new client configuration by clicking _Authorize_.
+
+**OAuth**
+
+To create a [OAuth 2.0](https://developers.google.com/identity/protocols/oauth2/web-server) integration:
+
+3. Still under **APIs and Services**, in **Credentials**, create a new **OAuth client ID** selecting the **Web application** application type.
```
glennmatthews, line 133: Does this generate the required credentials file? Step 3 under "Service Account" mentions saving a credentials file, but this step here does not.

nautobot_circuit_maintenance/models.py
chadell: totally, it's more pythonic and also moved `save()` to the caller.

glennmatthews (approved):

PR #42, opened 2021-07-21
Error handling and logging improvements

glennmatthews:
* Improve the wording and formatting of various job log messages
* Improve listing of configured providers in notificationsource detail view
* Avoid erroring out when encountering a Provider that has a null `emails_circuit_maintenances` custom field value
* Handle case where `emails_circuit_maintenances` comma-separated list contains whitespace

### DIFF
--- a/nautobot_circuit_maintenance/handle_notifications/sources.py
+++ b/nautobot_circuit_maintenance/handle_notifications/sources.py
@@ -6,6 +6,7 @@ import re
 import datetime
 import email
 import json
+import traceback
 from urllib.parse import urlparse
 from email.utils import mktime_tz, parsedate_tz
 from typing import Iterable, Optional, TypeVar, Type, Tuple, Dict, Union
@@ -624,8 +625,9 @@ def get_notifications(
                     )
 
         except Exception as error:
-            job_logger.log_warning(
-                message=f"Issue fetching notifications from {notification_source.name}: {error}",
+            stacktrace = traceback.format_exc()
+            job_logger.log_failure(
+                message=f"Issue fetching notifications from {notification_source.name}: {error}\n```\n{stacktrace}\n```",
             )
 
     return received_notifications
--- a/nautobot_circuit_maintenance/handle_notifications/sources.py
+++ b/nautobot_circuit_maintenance/handle_notifications/sources.py
@@ -332,7 +332,7 @@ class IMAP(EmailSource):
         source_header = settings.PLUGINS_CONFIG["nautobot_circuit_maintenance"]["source_header"]
         email_source = self.extract_email_source(email_message[source_header])
         if not email_source:
-            job_logger.log_failure(
+            job_logger.log_warning(
                 message="Not possible to determine the email sender from "
                 f'"{source_header}: {email_message[source_header]}"'
             )
@@ -487,7 +487,7 @@ class GmailAPI(EmailSource):
         email_source_before = email_source
         email_source = self.extract_email_source(email_source)
         if not email_source:
-            job_logger.log_failure(message=f"Not possible to determine the email sender: {email_source_before}")
+            job_logger.log_warning(message=f'Not possible to determine the email sender from "{email_source_before}"')
             return None
 
         provider_data_types, provider_type, error_message = self.extract_provider_data_types(email_source)
@@ -561,7 +561,7 @@ class GmailAPI(EmailSource):
             if raw_notification:
                 received_notifications.append(raw_notification)
 
-        job_logger.log_debug(message=f"Raw notifications: {received_notifications}")
+        # job_logger.log_debug(message=f"Raw notifications: {received_notifications}")
 
         self.close_service()
         return received_notifications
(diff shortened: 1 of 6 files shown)


dgarros (approved):
LGTM

nautobot_circuit_maintenance/handle_notifications/sources.py
```diff
@@ -253,8 +254,10 @@ def extract_provider_data_types(email_source: str) -> Tuple[str, str, str]:
             )
         """
         for provider in Provider.objects.all():
-            if "emails_circuit_maintenances" in provider.custom_field_data:
-                if email_source in provider.custom_field_data["emails_circuit_maintenances"].split(","):
+            if provider.custom_field_data.get("emails_circuit_maintenances"):
```
itdependsnetworks, line 257: Why not in single if conditional?

nautobot_circuit_maintenance/tests/test_sources.py
```diff
@@ -128,7 +128,7 @@ def test_extract_provider_data_types_no_provider_parser(self):
             (
                 "",
                 "",
-                f"Unexpected provider {provider_type} received from {email_source}, so not getting the notification",
+                f"{email_source} is for provider type `{provider_type}`, not presently supported",
```
itdependsnetworks, line 131:
```suggestion
                f"{email_source} is for provider type `{provider_type}`, which is not presently supported",
```

nautobot_circuit_maintenance/tests/test_sources.py
```diff
@@ -152,7 +152,7 @@ def test_extract_provider_data_types_ok(self, provider_type, data_types, error_m
             (
                 data_types if not error_message else "",
                 provider_type if not error_message else "",
-                f"Unexpected provider unknown received from {email_source}, so not getting the notification"
+                f"{email_source} is for provider type `{provider_type}`, not presently supported"
```
itdependsnetworks, line 155:
```suggestion
                f"{email_source} is for provider type `{provider_type}`, which is not presently supported"
```

nautobot_circuit_maintenance/handle_notifications/sources.py
glennmatthews: You mean:

```python
             if (
                 provider.custom_field_data.get("emails_circuit_maintenances")
                 and email_source in [
                    source.strip() for source in provider.custom_field_data["emails_circuit_maintenances"].split(",")
                ]
            ):
```

Feels a bit unwieldy to me but I'm not opposed to it.

itdependsnetworks: i won't lose sleep over it and only see pros/cons. If I see dual if statements I am confused, conversely if I see complex if statement its hard to read.

Also thinking to invert the truth table to keep it less nested.

```
            if provider.custom_field_data.get("emails_circuit_maintenances") is None:
               continue
            if email_source in [
                    source.strip() for source in provider.custom_field_data["emails_circuit_maintenances"].split(",")
                ]:
                provider_type = provider.slug
                break
```

glennmatthews:
Addressed review comments - thanks! Some additional fixes and enhancements:

- Clarified some aspects of README.md
- Fixed an issue where we weren't recognizing email addresses containing `-` characters as valid
- Added `source_header` configuration option to support a case where messages are forwarded from a mailing list and so we need to inspect the `X-Original-From` header instead of the `From` header to identify the actual message source.

nautobot_circuit_maintenance/handle_notifications/sources.py
```diff
@@ -220,6 +221,7 @@ def validate_providers(self, job_logger: Job, notification_source: NotificationS
             )
             return False
 
+        # job_logger.log_debug(message=f"Fetching emails from {self.emails_to_fetch}")
```
chadell, line 224: is this left behind  for easy debugging when needed?

chadell (commented):
nice improvements!
My only concern is about the scope of `source_header` because it looks to me as provider specific, so the same instance of the Plugin could be parsing emails from a provider that uses the "From" and other that uses the "X-Original-From", and from the implementation we are only covering on or another.
So, naturally it comes to me to see this as an extra optional custom field for the Provider, that would use "From" if not specified.
And, as you already pointed out on the PR, we should take this into account when fetching emails from a Source, using the "source_header".

nautobot_circuit_maintenance/handle_notifications/sources.py
glennmatthews: Yeah - it was useful in troubleshooting why some emails weren't getting fetched, but is a bit noisy during normal operation. 

I see some other plugins have added an explicit "Debug" or "Verbose" checkbox to their Jobs that can be toggled to get more detailed logs when desired - maybe something to consider adding here?

glennmatthews:
I agree that the `source_header` should probably be a provider-specific configuration - I'll look into fixing that as soon as I can find the time.

nautobot_circuit_maintenance/handle_notifications/sources.py
chadell: yes, it would be great

chadell (approved):
LGTM

PR #44, opened 2021-07-22
Update .travis.yml

ubajze:
@chadell I have enabled GitHub auto-release in .travis.yml.

If there is a reason to comment out the `deploy-github` stage, let me know. Otherwise, you can merge this PR.

Uros


chadell (approved):
I remember some weird behaviour in the past when I was setting it up and forgot to activate again later because I was not using the Github releases.
I don't see a problem reactivating again.

Issue #46, opened 2021-07-26
"since" filter is too coarse-grained


glennmatthews:
### Environment

- Python version: 
- Nautobot version: 
- nautobot-circuit-maintenance version: 

### Expected Behavior

Running the job should pick up all emails that arrived since the last time the job was run

### Observed Behavior

Because the `since` filtering is date-based (not date-time based), any emails that arrived during the same day as the last job run will be skipped over, even though they haven't been processed yet.

### Steps to Reproduce

1.
2.
3.

glennmatthews:
Maybe filtering by read/unread status might be a viable alternative?

chadell:
I've been investigating this, and the goal of the SINCE filter was to limit the emails read to only do it once.
You are right on the observation about SINCE not being good enough for a periodic fetch of notifications, and I agree with you that a viable option, either for IMAP and Gmail API, would be to use the SEEN/UNSEEN flag to retrieve only the UNSEEN and to mark a email as SEEN after being processed.

chadell:
implemented in #53

PR #50, opened 2021-08-04
Replace SINCE filter by SEEN and UNSEEN approach

chadell:
This addresses issue #46 

The approach, as suggested in the issue, is to get rid fo the `SINCE` filter and use a `SEEN/UNSEEN` strategy for `IMAP` and `READ/UNREAD` for `GmailAPI`.

So, we fetch all the emails that are UNSEEN or UNREAD, and then we mark them as SEEN or remove the UNREAD label.

To make this possible, we had to get the extra SCOPE `modify` in Gmail API integration.


dgarros:
How would this solution work if we have multiple instances of the plugin connected to the same inbox ? 
For example, I can imagine having the production instance and the development instance using the same inbox. in this case would all emails be processed by both instances of the plugin or not ?

chadell:
As you have already foreseen, one email box should be only processed by one instance, because when the email is read, it's is marked as SEEN and other instances will skip it.
I don't see this as a big issue because the development scenario I imagine, each environment would have a different email box integration to not interfere, but if multiple Circuit Maintenance instances interacting with the same box is a requirement, then this is not a valid solution.

chadell:
closing this because with this approach, only one concurrent integration is possible

Issue #43, opened 2021-07-22
More flexible mapping of Provider (Nautobot) to Provider (circuit-maintenance-parser)


glennmatthews:
### Environment

- Nautobot version: 
- nautobot-circuit-maintenance version: 

### Proposed Functionality

Mapping Nautobot Providers to circuit-maintenance-provider Provider classes based on the Nautobot Provider `slug` is inflexible as it relies on enforcing that the Nautobot Provider `slug` be exactly equal to the circuit-maintenance-provider Provider class `__name__`. In practice this may not always be the case, and it's not always going to be feasible to change the existing `slug` in Nautobot to resolve this.

### Use Case

I have a Nautobot deployment with circuits defined for provider `CenturyLink` (former name of the provider now called Lumen). Without changing the `slug` of this provider, or creating a new `Lumen` provider and remapping all of the circuits to this new provider, I have no way to make emails from Lumen be handled by the Lumen parser provider because the plugin is looking for a parser called `centurylink`.

glennmatthews:
Maybe another optional custom field on Provider?

chadell:
Addressed in #51

PR #51, opened 2021-08-05
Flexible mapping provider to parser

chadell:
This PR addresses issue #43 

It creates a new custom field in `Provider` that can be used to define a custom Provider name in the `circuit-maintenance-parser` library, and when not provider, the `Provider.slug` would be used, same as today.


nautobot_circuit_maintenance/__init__.py
```diff
@@ -66,7 +71,7 @@ class CircuitMaintenanceConfig(PluginConfig):
 
     def ready(self):
         super().ready()
-        post_migrate.connect(custom_field_extension, sender=self)
+        post_migrate.connect(custom_fields_extension, sender=self)
```
glennmatthews, line 74: We really need a better pattern for the CF creation - as I don't think this signal will get re-triggered when upgrading to a new version of the plugin in isolation, since we don't have any new migrations to apply? I wonder if we should explicitly call `custom_fields_extension` from the circuits job as well so as to ensure that it's always up-to-date whenever the job gets run?

Ref: https://github.com/nautobot/nautobot/issues/13

chadell: I see your point... but if I understand it correctly, calling it from the CM Job it's late as it will try to use the field, and only after it would be available, so it mitigates the issue but doesn't solves it.

I've reviewed the available signals from [here](https://docs.djangoproject.com/en/3.2/ref/signals/), and I don't see any better one than `post_migrate`. For the long term, to solve the issue that you referenced, maybe Nautobot could implement some custom signal when installing that this could be connected to.

For the short term (this PR), would be a solution to create a fake migration (without changes), just to make sure that this is triggered, and we could adopt the new pattern when proposed in Nautobot 1.2. WDYT?

glennmatthews: I'm probably missing some nuance - why would it be too late to call it as a first step in `HandleCircuitMaintenanceNotifications.run()`?

chadell: Because the Job is using the content of the custom fields to customize the retrieval of notifications, so the it expects the value from these fields to be in place. 
My understanding is that the first time that the Job would be run, the Custom Fields would be created (if not done before from the post_migration signal), so then you could go to the `Provider` to update them, but the first execution of the Job would be useless

glennmatthews: Gotcha. The first execution of the job would follow the existing behavior (no flexible mapping) but after that first run it would be possible to add flexible mappings - which I think is still a decent fallback option, unless you can easily set up the necessary mock migration instead?

chadell: trying 🥷 mode

chadell: bf8befc worked like a charm (not very proud of me)
I've tested the migration path and works fine

glennmatthews (dismissed):
Probably could use an update to README.md to explain this new functionality? Otherwise looks good!

chadell:
@glennmatthews I'm gonna merge this as it's already approved and 3985590 only updated the Readme and Changelog

Issue #48, opened 2021-07-27
Rendering of `raw` field in a RawNotification detail view could render text


glennmatthews:
### Environment

- Nautobot version: 1.0.3
- nautobot-circuit-maintenance version: 0.1.7

### Proposed Functionality

In the common case where a RawNotification's `raw` data is actually a human-readable string, it would be nice to render it as such, instead of as a raw byte-string. I.e., it currently appears like:

![image](https://user-images.githubusercontent.com/5603551/127157662-f2d79c7f-d81b-47c4-9aee-fff838c651e0.png)

but it would be nicer to render the text, for example as:

![image](https://user-images.githubusercontent.com/5603551/127157927-0f60aac4-6f8c-4b45-a5ec-85f0238f0d13.png)

chadell:
Implemente in #55

PR #55, opened 2021-08-09
Alter raw from RawNotifications to become a BinaryField and improve field rendering

chadell:
- Addresses #48: Simplifies usage of `raw` argument as `BinaryField`. Using a complex migration in order to avoid issues converting directly from string to binary in postgres.


nautobot_circuit_maintenance/migrations/0007_raw_binary.py
```diff
@@ -0,0 +1,44 @@
...
+
+def convert_raw_from_str_to_bytes(apps, schema_editor):
+    """Convert raw from str to bytes."""
+    RawNotificationModel = apps.get_model("nautobot_circuit_maintenance", "RawNotification")
+
+    for raw_notification in RawNotificationModel.objects.all():
+        raw_txt = raw_notification.raw
+        raw_txt = raw_txt.strip("b''").strip("''")
+        raw_notification.raw_b = raw_txt.encode()
+        raw_notification.save(update_fields=["raw_b"])
```
glennmatthews, line 14: So if I understand this right, we essentially have a bug in the previous version in that we were storing a quoted string representation of the raw bytes in a TextField, and now we're trying to migrate to actually storing the raw bytes correctly in a BinaryField? I'm pretty sure that stripping the leading `b'` and trailing `'` and then encoding it to utf-8 are *not* going to give the desired results. If you have a local instance stood up with some of the bad data, I'm happy to jump into nbshell and see if I can help figure out how we should address this - in the worst case, we're early enough yet that it might be easiest to simply throw away bad raw data as part of this migration.

nautobot_circuit_maintenance/views.py
```diff
@@ -204,7 +204,11 @@ def get_extra_context(self, request, instance):
             parsed_notification = models.ParsedNotification.objects.filter(raw_notification=instance).last()
         else:
             parsed_notification = None
-        return {"parsed_notification": parsed_notification}
+        raw_bytes = instance.raw.tobytes()
+        # TODO: Using unicode-escape to handle bytes converted from string from migration 0007
+        # This can be removed when squashing migrations on release 1.0.0
+        raw_repr = raw_bytes.decode("unicode-escape")
```
glennmatthews, line 210: This also seems wrong to me - shouldn't `instance.raw` already be `bytes`? Why do we need to call `tobytes` on it? I also don't like seeing "temporary" code here - it's not clear to me on the comment why this is needed now but won't be later?

chadell: instance.raw is a `memoryview` that I'm converting to `bytes` above.
the "temporary" code is because for the converted notifications when migration from Text to Binary, the escaped characters are converted from `\n` to `\\n`. If we found a better way to do the conversion we could skip this "unicode-escape" and leave as `decode()`

nautobot_circuit_maintenance/migrations/0007_raw_binary.py
chadell: yes, I was also considering the option of throw away bad raw data... but I will give an extra try to this conversion and if not I would ping you

nautobot_circuit_maintenance/views.py
```diff
@@ -204,7 +204,8 @@ def get_extra_context(self, request, instance):
             parsed_notification = models.ParsedNotification.objects.filter(raw_notification=instance).last()
         else:
             parsed_notification = None
-        return {"parsed_notification": parsed_notification}
+        raw_repr = instance.raw.tobytes().decode()
```
glennmatthews, line 207: Do we need exception handling for a potential case where `instance.raw` is *not* a binary-encoded string? Say if it's some arbitrary blob of binary data like (shudder) a PNG file?

glennmatthews:
Can you provide a screenshot showing how a newly imported (correct `raw` format) notification gets rendered in the web view? Do line breaks get rendered correctly now?

nautobot_circuit_maintenance/views.py
chadell: the data is gathered from email parts of types: "text/calendar" and "text/html". but adding an exception could save some future issue

nautobot_circuit_maintenance/views.py
```diff
@@ -204,7 +204,12 @@ def get_extra_context(self, request, instance):
             parsed_notification = models.ParsedNotification.objects.filter(raw_notification=instance).last()
         else:
             parsed_notification = None
-        raw_repr = instance.raw.tobytes().decode()
+        try:
+            raw_repr = instance.raw.tobytes().decode("utf-8", "strict")
+        except UnicodeDecodeError as exc:
+            raw_repr = "Raw content was not able to be decoded with utf-8"
+            logger.warning(raw_repr + ": %s", exc)
```
glennmatthews, line 211:
```suggestion
            logger.warning("%s: %s", raw_repr, exc)
```

glennmatthews:
Can you provide a screenshot showing correct rendering of the raw notification in the UI now? I'm lazy and don't want to spend the time downloading this and setting it up to re-ingest a sample notification from Gmail just to test this out. :-)

chadell:
# Migrated notifications:
image

# New notifications
image


I notice that the render is not good for "migrated" notifications. should this be okish?

glennmatthews (approved):

PR #67, opened 2021-08-12
Release v0.1.9

chadell:


chadell:
> I'm not necessarily opposed to this, but it also shouldn't be necessary for us to release a new version of this plugin every time there's a patch version increment of the circuit-maintenance-parser. `^1.2.2` includes implicit compatibility with all versions up to but not including `2.0`. We should only need to do this if we're explicitly _dropping_ compatibility with an earlier version of the parser.

yes, I tried to be more explicit on the change, and actually in this case version `1.2.2` is not properly working, so I wanted to skip it, but I could just skip this one with support for previous versions.

glennmatthews (approved):

Issue #72, opened 2021-08-18
integrate with capacity metrics plugin to expose the maintenance status of a circuit as a metric


dgarros:
### Proposed Functionality
It would be very useful to integrate with the [capacity metrics plugin](https://github.com/nautobot/nautobot-plugin-capacity-metrics) to expose the maintenance status of a circuit as a metric. 

The goal is to expose a metric that will reflect the maintenance status of each circuit, something similar to Admin Status and Oper status for interfaces
- `1` circuit is operational
- `2` circuit is in maintenance mode

```
# Circuit operational
circuit_maintenance_status{"circuit": "XXXXX", provider="ntt", circuit_type="peering", site='XX"} 1.0

# Circuit in maintenance mode
circuit_maintenance_status{"circuit": "YYYYYY", provider="ntt", circuit_type="peering", site='YY"} 2.0
```

This is something we are already doing with (few) other plugins, here is an example for the chatops plugin
https://github.com/nautobot/nautobot-plugin-chatops/blob/develop/nautobot_chatops/__init__.py#L59

### Use Case
I can think of 2 main use cases for this feature:
- Visualize in Grafana the maintenance status of a circuit overtime.
- Identify when a circuit is in maintenance to avoid generating an alert, from the alertmanager for example.

chadell:
while working on it, I would only implement the "maintenance mode" as the other is just business as usual:
```
# HELP nautobot_circuit_maintenance_total Active Circuit Maintenances per circuit
# TYPE nautobot_circuit_maintenance_total counter
nautobot_circuit_maintenance_total{circuit="1111111",impact="OUTAGE",circuit_maintenance="123412341234",circuit_type="asdfgfdg",provider="ntt",site="site to define"} 1.0
```

chadell:
Addressed in #73

PR #73, opened 2021-08-20
Implement Circuit Maintenance metrics

chadell:
Addresses #72 

- [x] Tests
- [x] Changelog
- [x] Documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,7 +206,7 @@ Attributes:
 
 ### Metrics
 
-Leveraging on `nautobot-capacity-metrics` plugin, the `nautobot-circuit-maintenance` plugin will expose application metrics at `/api/plugins/capacity-metrics/app-metrics`
+Leveraging the `nautobot-capacity-metrics` plugin, the `nautobot-circuit-maintenance` plugin can expose application metrics at `/api/plugins/capacity-metrics/app-metrics` if desired.
 
 Current exposed metric is the `circuit operational status` which shows the operational status for each `Circuit`(attached to a `CircuitTermination`) depending on related Circuit Maintenances (1: Operational, 0: Under active maintenance):
 
@@ -219,7 +219,7 @@ circuit_maintenance_status{circuit="2222222",circuit_type="Peering",provider="co
 circuit_maintenance_status{circuit="2222222",circuit_type="Peering",provider="colt",site="Girona",status="active-maintenance"} 0.0
 ```
 
-Metric generation is **disabled** by default, to **enable** them, add a `enable: True` in the `nautobot_circuit_maintenance.metrics` dict.
+Metric generation is **disabled** by default, to **enable** them, add a `enable: True` in the `nautobot_circuit_maintenance.metrics` dict. (Of course you must also install the `nautobot_capacity_metrics` plugin and ensure that it is included in `PLUGINS` as a prerequisite to enabling this feature.)
 
 By default, each circuit has attached some labels and values (cid, provider, type and site), but these labels can be defined in the Plugin configuration by adding an optional dictionary (under "metrics" -> "labels_attached") with the label name and the attributes within the Circuit object. (Note: in case of a value that can be multiple values, such as `terminations`, the first defined one will be used)
 
--- a/README.md
+++ b/README.md
@@ -208,13 +208,15 @@ Attributes:
 
 Leveraging on `nautobot-capacity-metrics` plugin, the `nautobot-circuit-maintenance` plugin will expose application metrics at `/api/plugins/capacity-metrics/app-metrics`
 
-Current exposed metric is the `circuit operational status` which shows the operational status for each `Circuit`(attached to a `CircuitTermination`) depending on related Circuit Maintenances (1: Operational, 2: Under active maintenance):
+Current exposed metric is the `circuit operational status` which shows the operational status for each `Circuit`(attached to a `CircuitTermination`) depending on related Circuit Maintenances (1: Operational, 0: Under active maintenance):
 
 ```
-# HELP nautobot_circuit_maintenance_total Circuit operational status
-# TYPE nautobot_circuit_maintenance_total counter
-nautobot_circuit_maintenance_total{circuit="1111111",circuit_type="Transit",provider="ntt",site="Barcelona"} 2.0
-nautobot_circuit_maintenance_total{circuit="2222222",circuit_type="Peering",provider="colt",site="Girona"} 1.0
+# HELP circuit_maintenance_status Circuit operational status
+# TYPE circuit_maintenance_status gauge
+circuit_maintenance_status{circuit="1111111",circuit_type="Transit",provider="ntt",site="Barcelona",status="operational"} 0.0
+circuit_maintenance_status{circuit="1111111",circuit_type="Transit",provider="ntt",site="Barcelona",status="active-maintenance"} 1.0
+circuit_maintenance_status{circuit="2222222",circuit_type="Peering",provider="colt",site="Girona",status="operational"} 1.0
+circuit_maintenance_status{circuit="2222222",circuit_type="Peering",provider="colt",site="Girona",status="active-maintenance"} 0.0
 ```
 
 Metric generation is **disabled** by default, to **enable** them, add a `enable: True` in the `nautobot_circuit_maintenance.metrics` dict.
--- a/README.md
+++ b/README.md
@@ -208,7 +208,7 @@ Attributes:
 
 Leveraging the `nautobot-capacity-metrics` plugin, the `nautobot-circuit-maintenance` plugin can expose application metrics at `/api/plugins/capacity-metrics/app-metrics` if desired.
 
-Current exposed metric is the `circuit operational status` which shows the operational status for each `Circuit`(attached to a `CircuitTermination`) depending on related Circuit Maintenances (1: Operational, 0: Under active maintenance):
+Current exposed metric is the `circuit operational status` which shows the operational status for each `Circuit`(attached to a `CircuitTermination`) depending on related Circuit Maintenances (1: Operational, 2: Under active maintenance):
 
 ```
 # HELP circuit_maintenance_status Circuit operational status
--- a/README.md
+++ b/README.md
@@ -213,10 +213,8 @@ Current exposed metric is the `circuit operational status` which shows the opera
 ```
 # HELP circuit_maintenance_status Circuit operational status
 # TYPE circuit_maintenance_status gauge
-circuit_maintenance_status{circuit="1111111",circuit_type="Transit",provider="ntt",site="Barcelona",status="operational"} 0.0
-circuit_maintenance_status{circuit="1111111",circuit_type="Transit",provider="ntt",site="Barcelona",status="active-maintenance"} 1.0
-circuit_maintenance_status{circuit="2222222",circuit_type="Peering",provider="colt",site="Girona",status="operational"} 1.0
-circuit_maintenance_status{circuit="2222222",circuit_type="Peering",provider="colt",site="Girona",status="active-maintenance"} 0.0
+circuit_maintenance_status{circuit="1111111",circuit_type="Transit",provider="ntt",site="Barcelona"} 2.0
+circuit_maintenance_status{circuit="2222222",circuit_type="Peering",provider="colt",site="Girona"} 1.0
 ```
 
 Metric generation is **disabled** by default, to **enable** them, add a `enable: True` in the `nautobot_circuit_maintenance.metrics` dict. (Of course you must also install the `nautobot_capacity_metrics` plugin and ensure that it is included in `PLUGINS` as a prerequisite to enabling this feature.)
--- a/README.md
+++ b/README.md
@@ -208,12 +208,13 @@ Attributes:
 
 Leveraging on `nautobot-capacity-metrics` plugin, the `nautobot-circuit-maintenance` plugin will expose application metrics at `/api/plugins/capacity-metrics/app-metrics`
 
-Current exposed metric is the `Active Circuit Maintenances per circuit` which shows all the circuits that have an active maintenance:
+Current exposed metric is the `ircuit operational status` which shows the operational status for each Circuit depending on related Circuit Maintenances (1: Operational, 2: Under active maintenance):
 
 ```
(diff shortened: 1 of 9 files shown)


dgarros (commented):
Thank you for starting this effort. 

For the use cases that I had in mind, the idea was to generate 1 metric per circuit, not per circuit maintenance, or maybe one metrics per circuit termination in order to include the site and the region. 

The 1 metric per circuit would represent the expected status of the circuit, has a maintenance active or does not, and this way it would allow us to leverage this metric to filter alerts. 

In some way we could just expose the existing status of a circuit as metric (and we should) but since we don't have a way currently to automatically update the status of the circuit when a maintenance is starting / finishing I think our best option is to dynamically generate this metric based on the information we have in the database.

Also in this case it's important to always generate a metric per circuit, not just when the circuit is in maintenance.

I hope this make sense, happy to discuss more about it next week

nautobot_circuit_maintenance/__init__.py
```diff
@@ -82,5 +82,11 @@ def ready(self):
         post_migrate.connect(custom_fields_extension, sender=self)
         post_migrate.connect(import_notification_sources, sender=self)
 
+        # pylint: disable=import-outside-toplevel
+        from nautobot_capacity_metrics import register_metric_func
+        from .metrics_app import metric_active_maintenances
+
+        register_metric_func(metric_active_maintenances)
```
dgarros, line 89: I think it would be good to have a flag in the config of the plugin to enable / disable this feature if needed

mzbroch:
Although we stanadrize labels, would be nice to have the label / values as a configuration option defined in settings:

```
labels_attached = {
   cid = circuit.cid,
   site = circuit.circuit_termination.site.name,
   etc
}
```

chadell:
@mzbroch , I took into account your comment in https://github.com/nautobot/nautobot-plugin-circuit-maintenance/pull/73/commits/f51063bbf6b78c368fca379939757156db5ca2f0
I discarded other options such as creating a `get_circuit_labels` method in `nautobot_config.py`  to avoid moving some complexity to handle some cases (such as `terminations`) into the user side, and leaving it the simplest as possible.
Open to feedback

nautobot_circuit_maintenance/metrics_app.py
```diff
@@ -0,0 +1,89 @@
...
+    # Statuses that we understand a Circuit Maintenance is expected to run
+    # Not all the providers use all the standard statuses.
+    active_statuses = ["CONFIRMED", "IN-PROCESS", "RE-SCHEDULED"]
+    active_circuit_maintenances = CircuitMaintenance.objects.filter(
+        status__in=active_statuses, start_time__lte=datetime.utcnow(), end_time__gte=datetime.utcnow()
+    )
+
+    for circuit in Circuit.objects.all():
+        status = 1
+        if CircuitImpact.objects.filter(circuit=circuit, maintenance__in=active_circuit_maintenances).exclude(
```
dgarros, line 70: I think this probably won't scale if we have 100s or 1000s of circuits because it will make 1 database query per Circuit. 
this function will be called very frequently so we need the response time to be as small as possible
We need to find a way to pull all the information we need in as few requests as possible and if we have to it's probably best to do some processing in Python here vs executing too many DB queries

dgarros (commented):
Almost there, thanks for making the suggested changes

nautobot_circuit_maintenance/metrics_app.py
```diff
@@ -0,0 +1,89 @@
...
+    )
+
+    # Statuses that we understand a Circuit Maintenance is expected to run
+    # Not all the providers use all the standard statuses.
+    active_statuses = ["CONFIRMED", "IN-PROCESS", "RE-SCHEDULED"]
+    active_circuit_maintenances = CircuitMaintenance.objects.filter(
+        status__in=active_statuses, start_time__lte=datetime.utcnow(), end_time__gte=datetime.utcnow()
+    )
+
+    for circuit in Circuit.objects.all():
```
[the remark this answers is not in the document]
chadell: I did a small trick here... taking the suggestion from Marek about leaving it to the user to define the labels/info, I implemented this recursive function `rgetattr` that covers this use case...
When targeting `"site": "terminations.site.name",`, when the `terminations` key is evaluated, it contains either Termination A or Z (there are also `termination_a` and `termination_z`, but then we would miss some cases), and checks for a valid Termination to take the data from and return.

nautobot_circuit_maintenance/metrics_app.py
chadell: very good point!, I will work on some `prefetch` work to decrease the query time.

nautobot_circuit_maintenance/metrics_app.py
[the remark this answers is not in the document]
dgarros: How would it work if both `termination_a` and `termination_z` are defined, it's common for backbone links between sites

dgarros (dismissed):
Looks great, I'm looking forward to seeing this one deploy in production

nautobot_circuit_maintenance/metrics_app.py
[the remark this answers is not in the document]
chadell: Then, if a `Circuit` goes from `Site A` to `Site B`,  the value expected for `Site` should be `Site A, Site B`? 
or maybe better to have two labels such as `Site Termination A` = `Site A` and `Site Termination Z` = `Site B`?

dgarros (dismissed):
Looks great, thanks for the super fast turnaround on this feature

README.md
```diff
@@ -204,6 +204,40 @@ Attributes:
...
+Leveraging on `nautobot-capacity-metrics` plugin, the `nautobot-circuit-maintenance` plugin will expose application metrics at `/api/plugins/capacity-metrics/app-metrics`
+
+Current exposed metric is the `circuit operational status` which shows the operational status for each `Circuit`(attached to a `CircuitTermination`) depending on related Circuit Maintenances (1: Operational, 2: Under active maintenance):
+
+```
+# HELP nautobot_circuit_maintenance_total Circuit operational status
+# TYPE nautobot_circuit_maintenance_total counter
+nautobot_circuit_maintenance_total{circuit="1111111",circuit_type="Transit",provider="ntt",site="Barcelona"} 2.0
+nautobot_circuit_maintenance_total{circuit="2222222",circuit_type="Peering",provider="colt",site="Girona"} 1.0
+```
```
glennmatthews, line 218: Ignorant question - is this a common/useful pattern in Prometheus? I naively would have expected:

1) this to be a Gauge, not a Counter
2) The status to be a distinct label with a value for each possible state (with gauge value 0 or 1 accordingly)

I.e. something like:

```
# HELP circuit_maintenance_status Circuit operational status
# TYPE circuit_maintenance_status gauge
circuit_maintenance_status{circuit="1111111",circuit_type="Transit",provider="ntt",site="Barcelona",status="operational"} 0.0
circuit_maintenance_status{circuit="1111111",circuit_type="Transit",provider="ntt",site="Barcelona",status="active-maintenance"} 1.0
circuit_maintenance_status{circuit="2222222",circuit_type="Peering",provider="colt",site="Girona",status="operational"} 1.0
circuit_maintenance_status{circuit="2222222",circuit_type="Peering",provider="colt",site="Girona",status="active-maintenance"} 0.0
```

pyproject.toml
```diff
@@ -21,6 +21,7 @@ packages = [
 [tool.poetry.dependencies]
 python = "^3.6.1"
 nautobot = "*"
+nautobot-capacity-metrics = "*"
```
glennmatthews, line 24: Can this be an optional/extras dependency? Or does it need to be a required dependency even if metrics aren't in use?

nautobot_circuit_maintenance/__init__.py
```diff
@@ -82,5 +82,13 @@ def ready(self):
         post_migrate.connect(custom_fields_extension, sender=self)
         post_migrate.connect(import_notification_sources, sender=self)
 
+        # pylint: disable=import-outside-toplevel
+        from nautobot_capacity_metrics import register_metric_func
+        from .metrics_app import metric_circuit_operational
```
glennmatthews, line 87: Possibly move these inside the below `if` block? No need to import them if they're not going to be used.

nautobot_circuit_maintenance/metrics_app.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def metric_circuit_operational():
+    """Expose the operational state of Circuits with a CircuitTermination when a Maintenance is ongoing.
+
+    # Circuit operational
+    circuit_maintenance_status{"circuit": "XXXXX", provider="ntt", circuit_type="peering", site='XX"} 1.0
+
+    # Circuit in maintenance mode
+    circuit_maintenance_status{"circuit": "YYYYYY", provider="ntt", circuit_type="peering", site='YY"} 2.0
```
glennmatthews, line 40: Some inconsistency here and in the docs - what's the actual metric name? Is it `circuit_maintenance_status` (here), `nautobot_circuit_maintenance` (below), or `nautobot_circuit_maintenance_total` (README.md)?

README.md
chadell: yes, I also believe that gauge makes more sense as it is not incremental

nautobot_circuit_maintenance/__init__.py
chadell: exactly

nautobot_circuit_maintenance/metrics_app.py
```diff
@@ -0,0 +1,81 @@
...
+        ):
+            status = 0
+
+        values = []
+        for _, attr in labels.items():
+            try:
+                label_value = rgetattr(termination, attr)
+            except AttributeError:
+                label_value = "n/a"
+            values.append(label_value)
```
glennmatthews, line 74: Is "n/a" a standard value in Prometheus? I would have expected to either set `label_value` to an empty string or else omit it entirely from `values` if it doesn't apply.

nautobot_circuit_maintenance/tests/test_app_metrics.py
```diff
@@ -0,0 +1,48 @@
...
+            self.assertEqual(len(circuit_metric.samples), 2)
+            for sample in circuit_metric.samples:
+                self.assertEqual(sample.labels["provider"], self.provider.name)
+                self.assertEqual(sample.labels["circuit_type"], self.circuit_type.name)
+                if sample.labels["circuit"] == self.circuit.cid:
+                    self.assertEqual(sample.labels["site"], self.site.name)
+                    self.assertEqual(sample.value, 0)
+                elif sample.labels["circuit"] == self.circuit_2.cid:
+                    self.assertEqual(sample.labels["site"], self.site.name)
+                    self.assertEqual(sample.value, 1)
```
glennmatthews, line 48:
```suggestion
                    self.assertEqual(sample.value, 1)
                else:
                    self.fail("Unexpected circuit label {sample.labels["circuit"]}"
```

nautobot_circuit_maintenance/tests/test_app_metrics.py
```diff
@@ -0,0 +1,48 @@
...
+    """Test cases for ensuring application metric endpoint is working properly."""
+
+    def setUp(self):
+        """Setup objects to run the test."""
+        self.provider = Provider.objects.create(name="Provider 1", slug="provider-1")
+        self.circuit_type = CircuitType.objects.create(name="Circuit Type 1", slug="circuit-type-1")
+        self.circuit = Circuit.objects.create(cid="Circuit 1", provider=self.provider, type=self.circuit_type)
+        self.circuit_2 = Circuit.objects.create(cid="Circuit 2", provider=self.provider, type=self.circuit_type)
+        self.circuit_3 = Circuit.objects.create(cid="Circuit 3", provider=self.provider, type=self.circuit_type)
+        self.site = Site.objects.create(name="Site 1", slug="site-1")
```
glennmatthews, line 21: Consider using different `type`, `provider`, and/or `site` values for each circuit so as to be able to test that the correct values are returned for each metric? Right now by using the same values across all circuits you're missing out on testing for some possible classes of bugs.

README.md
```diff
@@ -204,6 +204,42 @@ Attributes:
...
+Current exposed metric is the `circuit operational status` which shows the operational status for each `Circuit`(attached to a `CircuitTermination`) depending on related Circuit Maintenances (1: Operational, 0: Under active maintenance):
+
+```
+# HELP circuit_maintenance_status Circuit operational status
+# TYPE circuit_maintenance_status gauge
+circuit_maintenance_status{circuit="1111111",circuit_type="Transit",provider="ntt",site="Barcelona",status="operational"} 0.0
+circuit_maintenance_status{circuit="1111111",circuit_type="Transit",provider="ntt",site="Barcelona",status="active-maintenance"} 1.0
+circuit_maintenance_status{circuit="2222222",circuit_type="Peering",provider="colt",site="Girona",status="operational"} 1.0
+circuit_maintenance_status{circuit="2222222",circuit_type="Peering",provider="colt",site="Girona",status="active-maintenance"} 0.0
+```
```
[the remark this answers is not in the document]
chadell: I missed you point. I understand that the `status` is the status of the `Circuit`? 
but I still don't see why do we need two metrics by each `CircuitTermination` if it only can have one `status` and one gauge at a time... 🤔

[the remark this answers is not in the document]
glennmatthews: My thinking was that there might be more than two possible `status` values, so we'd need a different binary on/off metric for each possible status. If the only ever possible `status` values are just the two of `operational` and `active-maintenance`, then a single binary gauge is fine. Hope that makes sense.

README.md
dgarros: Gauge make sense but I would prefer to keep the value as 1 & 2 if possible, it's more inline what the industry is doing

README.md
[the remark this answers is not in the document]
dgarros: If we have more than 2 status we should use more numbers, for example to track the status of a BGP session the RFC is tracking 6 values
```
bgpPeerState OBJECT-TYPE
SYNTAX INTEGER {
idle(1),
connect(2),
active(3),
opensent(4),
openconfirm(5),
established(6)
}
```

glennmatthews (approved):
Taking @dgarros at his word and greater experience here that this is the logical and expected structure for the metrics. :-)

PR #78, opened 2021-08-31
Add support for `extra_scopes` on Gmail notification sources

glennmatthews:
I have a custom plugin that builds atop nautobot-circuit-maintenance to automatically update a Google Calendar when a CircuitMaintenance is created or updated. Rather than setting up a separate OAuth2 integration for Google Calendar, it'd be really useful to just piggyback atop the existing NotificationSource OAuth setup. Towards this end, this PR adds an optional `extra_scopes` plugin config item that can be used to define additional access scopes to request when setting up OAuth. I've tested it in our user's environment and it meets my needs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## v0.1.10
 
+### Added
+
+- #78: Add optional `extra_scopes` config parameter to use with Gmail notification sources.
+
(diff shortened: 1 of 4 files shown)


chadell (approved):
LGTM

chadell (approved):
LGTM

Issue #75, opened 2021-08-27
CircuitMaintenance "name" should probably include provider as well as maintenance_id


glennmatthews:
### Environment

- Nautobot version: 
- nautobot-circuit-maintenance version: 

### Proposed Functionality

As and when this plugin is used with an ever-growing set of providers sending maintenance notifications, it becomes increasingly likely that two different providers will send maintenance notifications that happen to use the same maintenance ID number/string. (Consider the case of two providers each sending their very first maintenance notification, and both choosing to identify it as maintenance id "00000001"). Because we use the `maintenance_id` as the globally-unique `name` of the `CircuitMaintenance` model instances, these two disparate sets of maintenance notifications would inadvertently be merged into a single `CircuitMaintenance` instance named `000000001`, which is almost certainly undesirable.

To avoid collision between providers in this fashion, we should probably include the notification `provider` value in addition to the `maintenance_id` in constructing the CircuitMaintenance `name`.

chadell:
@glennmatthews +1

chadell:
Addressed in #81

PR #81, opened 2021-09-03
Issue 75 avoid maintenance id collisions

chadell:
This PR is addressing 2 issues:
* #75 So the maintenance_id in the plugin now it's a composition of the `f"{provider.slug}-{maintenance_id}"
* The usage of custom mapping towards the parser provider had an issue to parse the notification

I'm aware of the lack of test for this second point that has been tested manually, but I plan to refactor the `source` and `handler` code with the new `parser` library and I will add this test then (will create an issue: #82 ), because the current tests are getting too complex (parallel to the code)

### DIFF
--- a/nautobot_circuit_maintenance/handle_notifications/handler.py
+++ b/nautobot_circuit_maintenance/handle_notifications/handler.py
@@ -151,7 +151,7 @@ def update_circuit_maintenance(
 
 def process_parsed_notification(logger: Job, parsed_notification: ParsedNotification, raw_entry: RawNotification):
     """Processes a Parsed Notification, creating or updating the related Circuit Maintenance."""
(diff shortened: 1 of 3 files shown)


nautobot_circuit_maintenance/handle_notifications/handler.py
```diff
@@ -185,9 +185,15 @@ def process_raw_notification(  # pylint: disable=too-many-branches
         )
         return None
 
+    provider_type = (
+        provider.get_custom_fields().get(CustomField.objects.get(name="provider_parser_circuit_maintenances"))
```
glennmatthews, line 189: I *think* this could be implemented more simply as `provider.cf.get("provider_parser_circuit_maintenances")`?

glennmatthews (approved):

Issue #84, opened 2021-09-06
 Change `NotificationSource` config between restarts cause DB errors


chadell:
### Environment

- Python version: 3.6
- Nautobot version: 1.1.2
- nautobot-circuit-maintenance version: 0.1.10

### Expected Behavior

When changing the `NotificationSource` config between Nautobot restarts nothing wrong should happen, just update as defined.

### Observed Behavior

When changing the configr from `PLUGINS_CONFIG` DB errors are raised:
```
nautobot_1       | nautobot_circuit_maintenance.models.DoesNotExist: NotificationSource matching query does not exist.
nautobot_1       | 
nautobot_1       | During handling of the above exception, another exception occurred:
nautobot_1       | 
nautobot_1       | Traceback (most recent call last):
nautobot_1       |   File "/usr/local/lib/python3.6/site-packages/django/db/backends/utils.py", line 84, in _execute
nautobot_1       |     return self.cursor.execute(sql, params)
nautobot_1       | psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "nautobot_circuit_maintenance_notificationsource_name_key"
nautobot_1       | DETAIL:  Key (name)=(my imap source) already exists.
```

### Steps to Reproduce

1.
2.
3.

chadell:
Fixed in #85

PR #85, opened 2021-09-06
Issue #84: Fix how attach_all_providers is updated from config

chadell:


nautobot_circuit_maintenance/__init__.py
```diff
@@ -49,8 +49,9 @@ def import_notification_sources(sender, **kwargs):  # pylint: disable=unused-arg
             slug=slugify(
                 notification_source["name"],
             ),
-            attach_all_providers=notification_source.get("attach_all_providers", False),
         )
+        instance.attach_all_providers = notification_source.get("attach_all_providers", False)
+        instance.save()
```
glennmatthews, line 54: Perhaps use `defaults` on `get_or_create` rather than a separate update and save step?

```python
        instance, _ = NotificationSource.objects.get_or_create(
            name=notification_source["name"],
            slug=slugify(
                notification_source["name"],
            ),
            defaults={
                "attach_all_providers": notification_source.get("attach_all_providers", False),
            }
        )
```

chadell: I've double checked and using `defaults` is not updating well the desired stated of `attach_all_providers` when it was already set.

glennmatthews: Ah that makes sense - need to change this to use `update_or_create` instead of `get_or_create`.

chadell: totally! (I was not aware of the `defaults` option, thanks)

glennmatthews (approved):

PR #86, opened 2021-09-08
New parser library

chadell:
Goal: Use new `circuit-maintenance-parser` library, with new methods to simply client usage and reorg code to simplify and increase test coverage

# Changes

- [x] Adopt the new `circuit-maintenance-parser` init methods, simplifying the logic to send the email data directly to the library. To homogenize the usage, in `GmailSources` we move to the `raw` data. This also allowed the reuse the `process_email` between `EmailSources`. This also removed a lot of logic regarding email manipulation.
- [x] Revisit the logic in `handler.py` to make it more readable and add new related testing. There are some changes on return objects in some functions (for instance, the `process_raw_notification` now creates the `RawNotification` even the provider _type is not supported by the parser library)
- [x] Check `Raw` visualisation and storage for `RawNotification` because now emails are fully dumped
- [x] Use `Stamp` from the parser `Maintenance` to set the `RawNotification.date`

I tried to simplify the logic as much as possible, but any suggestion on it will be more than welcome!

### DIFF
--- a/development/nautobot_config.py
+++ b/development/nautobot_config.py
@@ -241,26 +241,28 @@ PLUGINS = ["nautobot_circuit_maintenance"]
 # Each key in the dictionary is the name of an installed plugin and its value is a dictionary of settings.
 PLUGINS_CONFIG = {
     "nautobot_circuit_maintenance": {
-        "raw_notifications": {"initial_days_since": 100},
+        "raw_notifications": {"initial_days_since": 500},
         "notification_sources": [
             {
                 "name": "my imap source",
                 "account": os.environ.get("CM_NS_1_ACCOUNT", ""),
                 "secret": os.environ.get("CM_NS_1_SECRET", ""),
                 "url": os.environ.get("CM_NS_1_URL", ""),
-                "attach_all_providers": True,
+                # "attach_all_providers": True,
             },
             {
                 "name": "my gmail service account api source",
                 "url": os.environ.get("CM_NS_2_URL", ""),
                 "account": os.environ.get("CM_NS_2_ACCOUNT", ""),
                 "credentials_file": os.environ.get("CM_NS_2_CREDENTIALS_FILE", ""),
+                "attach_all_providers": True,
             },
             {
                 "name": "my gmail oauth api source",
                 "url": os.environ.get("CM_NS_3_URL", ""),
                 "account": os.environ.get("CM_NS_3_ACCOUNT", ""),
                 "credentials_file": os.environ.get("CM_NS_3_CREDENTIALS_FILE", ""),
+                # "attach_all_providers": True,
             },
         ],
         "metrics": {
--- a/nautobot_circuit_maintenance/handle_notifications/handler.py
+++ b/nautobot_circuit_maintenance/handle_notifications/handler.py
@@ -88,12 +88,12 @@ def update_circuit_maintenance(
 
     circuit_entries = CircuitImpact.objects.filter(maintenance=circuit_maintenance_entry)
 
-    parsed_cids = {parsed_circuit.circuit_id for parsed_circuit in parser_maintenance.circuits}
+    new_cids = {parsed_circuit.circuit_id for parsed_circuit in parser_maintenance.circuits}
     existing_cids = {circuit_entry.circuit.cid for circuit_entry in circuit_entries}
 
-    cids_to_update = parsed_cids & existing_cids
-    cids_to_create = parsed_cids - existing_cids
-    cids_to_remove = existing_cids - parsed_cids
+    cids_to_update = new_cids & existing_cids
+    cids_to_create = new_cids - existing_cids
+    cids_to_remove = existing_cids - new_cids
 
     for cid in cids_to_create:
         circuit_entry = Circuit.objects.filter(cid=cid, provider=provider.pk).last()
--- a/nautobot_circuit_maintenance/handle_notifications/handler.py
+++ b/nautobot_circuit_maintenance/handle_notifications/handler.py
@@ -167,23 +167,8 @@ def process_parsed_notification(
     logger.log_success(parsed_entry, message=f"Saved Parsed Notification for {maintenance_id}.")
 
 
-def process_raw_notification(  # pylint: disable=too-many-branches
-    logger: Job, notification: MaintenanceNotification
-) -> Union[int, None]:
-    """Processes a raw notification (maybe containing multiple parsed notifications).
-
-    It creates a RawNotification and if it could be parsed, create the corresponding ParsedNotification.
-    """
-    provider = Provider.objects.filter(slug=notification.provider_type).last()
-
-    if not provider:
-        logger.log_warning(
-            message=(
-                f"Raw notification not created because is referencing to a provider not existent: {notification.provider_type}"
-            )
-        )
-        return None
-
+def get_maintenances_from_notification(logger: Job, notification: MaintenanceNotification, provider: Provider):
+    """Use the `circuit_maintenance_parser` library to get Maintenances from the notification."""
     provider_type = provider.cf.get("provider_parser_circuit_maintenances", "").lower() or provider.slug
 
     parser_provider = init_provider(provider_type=provider_type)
@@ -195,7 +180,7 @@ def process_raw_notification(  # pylint: disable=too-many-branches
     data_to_process = init_data_email(notification.raw_payload)
 
     try:
-        parsed_notifications = parser_provider.get_maintenances(data_to_process)
+        return parser_provider.get_maintenances(data_to_process)
     except ProviderError:
         tb_str = traceback.format_exc()
         logger.log_info(message=f"Parsing failed for notification `{notification.subject}`:\n```\n{tb_str}\n```")
@@ -207,6 +192,12 @@ def process_raw_notification(  # pylint: disable=too-many-branches
         )
         return None
 
+
+def create_raw_nofitication(logger: Job, notification: MaintenanceNotification, provider: Provider):
+    """Create a RawNotification.
+
+    If it already exists, we return `None` to signal we are skipping it.
+    """
     # Insert raw notification in DB even failed parsing
     try:
         raw_entry, created = RawNotification.objects.get_or_create(
@@ -227,6 +218,31 @@ def process_raw_notification(  # pylint: disable=too-many-branches
         return None
 
     logger.log_success(raw_entry, message="Raw notification created.")
+
+    return raw_entry
+
+
+def process_raw_notification(logger: Job, notification: MaintenanceNotification) -> Union[int, None]:
+    """Processes a raw notification (maybe containing multiple parsed notifications).
+
+    It creates a RawNotification and if it could be parsed, create the corresponding ParsedNotification.
+    """
+    provider = Provider.objects.filter(slug=notification.provider_type).last()
+    if not provider:
+        logger.log_warning(
+            message=(
+                f"Raw notification not created because is referencing to a provider not existent: {notification.provider_type}"
+            )
+        )
+        return None
+
+    raw_entry = create_raw_nofitication(logger, notification, provider)
+    if not raw_entry:
+        return None
+
+    parsed_notifications = get_maintenances_from_notification(logger, notification, provider)
+    if not parsed_notifications:
+        return raw_entry.id
 
     for parsed_notification in parsed_notifications:
         try:
--- a/nautobot_circuit_maintenance/handle_notifications/handler.py
+++ b/nautobot_circuit_maintenance/handle_notifications/handler.py
@@ -318,7 +318,7 @@ class HandleCircuitMaintenanceNotifications(Job):
             )
         except Exception as error:
             self.log_failure(
-                message=f"Unexpected exception when retrieveing notifications from sources ({notification_sources}): {error}"
+                message=f"Unexpected exception when retrieving notifications from sources ({notification_sources}): {error}"
             )
             return []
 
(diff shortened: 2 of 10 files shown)


nautobot_circuit_maintenance/handle_notifications/handler.py
```diff
@@ -167,59 +167,43 @@ def process_parsed_notification(
...
-            return None
+    try:
+        return parser_provider.get_maintenances(data_to_process)
+    except ProviderError:
+        tb_str = traceback.format_exc()
+        logger.log_info(message=f"Parsing failed for notification `{notification.subject}`:\n```\n{tb_str}\n```")
+        return None
+    except Exception:
+        tb_str = traceback.format_exc()
+        logger.log_info(
```
glennmatthews, line 190: This feels like it should be `log_failure` to me - providers shouldn't be raising any Exceptions other than ProviderError, right?

nautobot_circuit_maintenance/handle_notifications/handler.py
```diff
@@ -167,59 +167,43 @@ def process_parsed_notification(
...
-    for raw_payload in notification.raw_payloads:
-        parser = init_provider(raw=raw_payload, provider_type=provider_type)
-        if not parser:
-            logger.log_warning(message=f"Notification Parser not found for {notification.provider_type}")
-            return None
+    try:
+        return parser_provider.get_maintenances(data_to_process)
+    except ProviderError:
+        tb_str = traceback.format_exc()
+        logger.log_info(message=f"Parsing failed for notification `{notification.subject}`:\n```\n{tb_str}\n```")
```
glennmatthews, line 186: This should probably at least be `log_warning`, and I could make an argument for `log_failure`. In integrating this plugin into an automated workflow (with daily or hourly runs of the importing job) we need to make it easy to automatically detect and flag parsing failures for telemetry, dashboards, etc.

nautobot_circuit_maintenance/handle_notifications/handler.py
```diff
@@ -167,59 +167,43 @@ def process_parsed_notification(
...
-    else:
-        parsed_notifications = []
-        logger.log_warning(message=f"Parsed failed for all the raw payloads for `{notification.subject}`.")
-        # Carry on with the last raw_payload in the list, if any
-        if not raw_payload:
-            return None
 
-    if isinstance(raw_payload, str):
-        raw_payload = raw_payload.encode("utf-8")
+def create_raw_nofitication(logger: Job, notification: MaintenanceNotification, provider: Provider):
```
glennmatthews, line 196:
```suggestion
def create_raw_notification(logger: Job, notification: MaintenanceNotification, provider: Provider):
```

nautobot_circuit_maintenance/handle_notifications/handler.py
```diff
@@ -235,6 +219,31 @@ def process_raw_notification(  # pylint: disable=too-many-branches
...
+    provider = Provider.objects.filter(slug=notification.provider_type).last()
+    if not provider:
+        logger.log_warning(
+            message=(
+                f"Raw notification not created because is referencing to a provider not existent: {notification.provider_type}"
+            )
+        )
+        return None
+
+    raw_entry = create_raw_nofitication(logger, notification, provider)
```
glennmatthews, line 239:
```suggestion
    raw_entry = create_raw_notification(logger, notification, provider)
```

nautobot_circuit_maintenance/handle_notifications/handler.py
```diff
@@ -235,6 +219,31 @@ def process_raw_notification(  # pylint: disable=too-many-branches
 
     logger.log_success(raw_entry, message="Raw notification created.")
 
+    return raw_entry
+
+
+def process_raw_notification(logger: Job, notification: MaintenanceNotification) -> Union[int, None]:
```
glennmatthews, line 225:
```suggestion
def process_raw_notification(logger: Job, notification: MaintenanceNotification) -> Optional[UUID]:
```

nautobot_circuit_maintenance/tests/test_handler.py
```diff
@@ -193,18 +201,18 @@ def test_run_invalid_notification(self):
     def test_process_raw_notification_no_provider_in_parser(self):
         """Test process_raw_notification with non existant Proivder in the parser library."""
         notification_data = get_base_notification_data()
-        test_notification = generate_raw_notification(notification_data, self.source.name)
+        test_notification = generate_email_notification(notification_data, self.source.name)
         test_notification.provider_type = "abc"
         res = process_raw_notification(self.job, test_notification)
-        self.assertEqual(res, None)
+        self.assertNotEqual(res, None)
```
glennmatthews, line 207: Why did this change? Are we now successfully processing the raw notification even though it's from an unknown provider_type?

chadell: I'm still thinking on the final logic, this was just a first try on the change, but the `process_raw_notification` now creates the `RawNotification` even it can't be parsed (in this case because the parser library has no `provider`) 
Obviously this is not the final word, still in `draft` mode

nautobot_circuit_maintenance/handle_notifications/handler.py
chadell: it took me 30 seconds to spot the typo 😄

nautobot_circuit_maintenance/models.py
```diff
@@ -14,6 +15,9 @@
 
 from .choices import CircuitImpactChoices, CircuitMaintenanceStatusChoices, NoteLevelChoices
 
+PLUGIN_SETTINGS = settings.PLUGINS_CONFIG.get("nautobot_circuit_maintenance", {})
+MAX_RAW_NOTIFICATION_SIZE = 1000
```
glennmatthews, line 19: Misleading name? We can configure a higher value than this if desired, right? Perhaps `DEFAULT_RAW_NOTIFICATION_SIZE` would be better?

This only applies to the RawNotification, right? The CircuitMaintenance will still store the full text of the email?

nautobot_circuit_maintenance/handle_notifications/handler.py
```diff
@@ -23,19 +24,19 @@
 
 
 def create_circuit_maintenance(
-    logger: Job, maintenance_id: str, parsed_notification: ParsedNotification, provider: Provider
+    logger: Job, maintenance_id: str, parser_maintenance, provider: Provider
```
glennmatthews, line 27: What's the type of `parser_maintenance` now?

Same question on each function that has this as an argument below.

nautobot_circuit_maintenance/handle_notifications/handler.py
```diff
@@ -235,22 +219,75 @@ def process_raw_notification(  # pylint: disable=too-many-branches
...
         try:
-            process_parsed_notification(logger, parsed_notification, raw_entry, provider)
-            # Update raw notification as properly parsed
+            circuit_maintenance_entry = create_or_update_circuit_maintenance(
+                logger, parser_maintenance, raw_entry, provider
+            )
+            # Update raw notification as properly parsed and with the stamp time
             raw_entry.parsed = True
+            raw_entry.date = datetime.datetime.fromtimestamp(parser_maintenance.stamp)
             raw_entry.save()
```
glennmatthews, line 256: Can we do these updates atomically as part of `create_or_update_circuit_maintenance` instead of as a separate step? I have an extension plugin that's listening for the `post_save` signal for `CircuitMaintenance` and I think with these changes it'll be triggered twice instead of once.

nautobot_circuit_maintenance/tests/test_handler.py
```diff
@@ -238,51 +247,79 @@ def test_process_raw_notification_parser_issue(self):
...
+        self.assertEqual(1, len(parser_maintenances))
+
+    def test_get_maintenances_from_notification_wrong_data(self):
+        """Test get_maintenances_from_notification."""
+        notification_data = get_base_notification_data()
+        notification_data["status"] = "Non valid status"
+        test_notification = generate_email_notification(notification_data, self.source.name)
+        provider = Provider.objects.get(slug=test_notification.provider_type)
+        parser_maintenances = get_maintenances_from_notification(self.job, test_notification, provider)
+        self.assertEqual(parser_maintenances, None)
```
glennmatthews, line 265:
```suggestion
        self.assertIsNone(parser_maintenances)
```

nautobot_circuit_maintenance/tests/test_handler.py
```diff
@@ -238,51 +247,79 @@ def test_process_raw_notification_parser_issue(self):
...
+        self.job.log_failure.assert_called()
+
+    def test_get_maintenances_from_notification_non_existent_provider_in_parser(self):
+        """Test get_maintenances_from_notification."""
+        notification_data = get_base_notification_data()
+        test_notification = generate_email_notification(notification_data, self.source.name)
+        provider = Provider.objects.get(slug=test_notification.provider_type)
+        provider.cf["provider_parser_circuit_maintenances"] = "unkown_provider_in_parser"
+        parser_maintenances = get_maintenances_from_notification(self.job, test_notification, provider)
+        self.assertEqual(parser_maintenances, None)
```
glennmatthews, line 275:
```suggestion
        self.assertIsNone(parser_maintenances)
```

nautobot_circuit_maintenance/tests/test_sources.py
```diff
@@ -114,74 +114,84 @@ def test_extract_email_source(self, email_source, email_source_output):
...
 
-        for email_source in email_sources:
-            self.assertEqual(
-                EmailSource.extract_provider_data_types(email_source),
-                ({"text/calendar"}, "ntt", ""),
-            )
+        provider_type = source.get_provider_type_from_email("user@example.com")
+        self.assertEqual(provider_type, "abc-d")
+        provider_type = source.get_provider_type_from_email("unknown")
+        self.assertEqual(provider_type, None)
```
glennmatthews, line 194:
```suggestion
        self.assertIsNone(provider_type)
```

nautobot_circuit_maintenance/handle_notifications/handler.py
chadell: it's `Maintenance` from the parser library (I added a TODO comment in some place), so I'm waiting for the official release to update

nautobot_circuit_maintenance/handle_notifications/handler.py
chadell: because this is for `RawNotification` I intentionally put outside of the `CircuitMaintenance` manipulation. I guess this should affect the signal, no?

nautobot_circuit_maintenance/models.py
chadell: The `CircuitMaintenance` doesn't store the email, only the `RawNotification`, so here we are actually truncating the original notification in the DB, but I believe that we should add some limit to prevent against big email attachements we could receive.
Maybe we could define a lower level to get some minimum size?

nautobot_circuit_maintenance/handle_notifications/handler.py
glennmatthews: Ah, my mistake! Fine with me to update the raw entry separately.

nautobot_circuit_maintenance/handle_notifications/handler.py
```diff
@@ -23,19 +24,19 @@
 
 
 def create_circuit_maintenance(
-    logger: Job, maintenance_id: str, parsed_notification: ParsedNotification, provider: Provider
+    logger: Job, maintenance_id: str, parser_maintenance, provider: Provider
 ) -> CircuitMaintenance:
     """Handles the creation of a new circuit maintenance."""
     circuit_maintenance_entry = CircuitMaintenance.objects.create(
```
glennmatthews, line 30: I just realized in testing my extension plugin that `CircuitMaintenance.objects.create()` doesn't trigger the `pre_save` or `post_save` signals, since it doesn't actually call `save()`. 😞 Would you be willing to change this to something like:

```python
circuit_maintenance_entry = CircuitMaintenance(
   ...
)
circuit_maintenance_entry.save()
```

so that the save signals are properly triggered as I need them to be?

chadell: sure! it's the best timing 😄

glennmatthews: Thanks!

development/nautobot_config.py
```diff
@@ -241,26 +241,28 @@
...
                 "account": os.environ.get("CM_NS_2_ACCOUNT", ""),
                 "credentials_file": os.environ.get("CM_NS_2_CREDENTIALS_FILE", ""),
+                "attach_all_providers": True,
             },
             {
                 "name": "my gmail oauth api source",
                 "url": os.environ.get("CM_NS_3_URL", ""),
                 "account": os.environ.get("CM_NS_3_ACCOUNT", ""),
                 "credentials_file": os.environ.get("CM_NS_3_CREDENTIALS_FILE", ""),
+                # "attach_all_providers": True,
```
glennmatthews, line 265: Should any/all of the changes in this file be reverted?

nautobot_circuit_maintenance/handle_notifications/handler.py
```diff
@@ -260,50 +298,45 @@ class Meta:
...
         try:
             notifications = get_notifications(
                 job_logger=self,
                 notification_sources=notification_sources,
-                since=since_reference,
+                since=get_since_reference(self),
+            )
+        except Exception as error:
+            self.log_failure(
+                message=f"Unexpected exception when retrieveing notifications from sources ({notification_sources}): {error}"
```
glennmatthews, line 321:
```suggestion
                message=f"Unexpected exception when retrieving notifications from sources ({notification_sources}): {error}"
```

glennmatthews (approved):

development/nautobot_config.py
chadell: I believe these changes are good for development (I have old emails to check 😄 )

PR #90, opened 2021-09-15
Implemented CI workflow in GH Actions

ubajze:


ubajze:
FYI: When this PR is merged, we need to delete the old PyPI token.

chadell:
> FYI: When this PR is merged, we need to delete the old PyPI token.

Do you mean that you have already rotated it and updated in the secrets backend?

ubajze:
> > FYI: When this PR is merged, we need to delete the old PyPI token.
> 
> Do you mean that you have already rotated it and updated in the secrets backend?

Correct. Also tested publishing and it works. This PR is ready to be merged.

chadell (approved):
LGTM

Issue #89, opened 2021-09-15
Fix yaml files to pass the yamllint step.


ubajze:
The following is the list of errors, which need to be fixed to pass the yamllint step:

```
INVOKE_NAUTOBOT_CIRCUIT_MAINTENANCE_LOCAL=True invoke yamllint
./nautobot_circuit_maintenance/fixtures/source_gmail_api.yaml
  2:10      error    string value is not quoted with double quotes  (quoted-strings)
  3:7       error    string value is not quoted with double quotes  (quoted-strings)
  8:36      error    string value is not quoted with double quotes  (quoted-strings)
  9:11      error    string value is not quoted with double quotes  (quoted-strings)
  10:11     error    string value is not quoted with double quotes  (quoted-strings)
  17:10     error    string value is not quoted with double quotes  (quoted-strings)
  18:7      error    string value is not quoted with double quotes  (quoted-strings)
  23:11     error    string value is not quoted with double quotes  (quoted-strings)
  24:11     error    string value is not quoted with double quotes  (quoted-strings)
  26:10     error    string value is not quoted with double quotes  (quoted-strings)
  27:7      error    string value is not quoted with double quotes  (quoted-strings)
  34:15     error    string value is not quoted with double quotes  (quoted-strings)
  35:11     error    string value is not quoted with double quotes  (quoted-strings)
  41:10     error    string value is not quoted with double quotes  (quoted-strings)
  42:7      error    string value is not quoted with double quotes  (quoted-strings)
  49:15     error    string value is not quoted with double quotes  (quoted-strings)
  50:11     error    string value is not quoted with double quotes  (quoted-strings)
  56:10     error    string value is not quoted with double quotes  (quoted-strings)
  57:7      error    string value is not quoted with double quotes  (quoted-strings)
  65:9      error    string value is not quoted with double quotes  (quoted-strings)
  66:10     error    string value is not quoted with double quotes  (quoted-strings)
  67:7      error    string value is not quoted with double quotes  (quoted-strings)
  75:9      error    string value is not quoted with double quotes  (quoted-strings)

./nautobot_circuit_maintenance/fixtures/handle_notifications_job.yaml
  2:10      error    string value is not quoted with double quotes  (quoted-strings)
  3:7       error    string value is not quoted with double quotes  (quoted-strings)
  8:36      error    string value is not quoted with double quotes  (quoted-strings)
  9:11      error    string value is not quoted with double quotes  (quoted-strings)
[...]

PR #91, opened 2021-09-15
Issue #89: Enable yamllint

chadell:


.yamllint.yml
```diff
@@ -1,5 +1,8 @@
 ---
 extends: "default"
+ignore: |
+  /nautobot_circuit_maintenance/fixtures/
```
glennmatthews, line 4: So the only files we're linting at this point are the docker-compose files? To my mind those aren't very useful to lint, but linting the test fixtures would be more useful.

chadell: The fixtures were taken directly from the DB dump export (this is the reason they don't have quotes) and used just for testing, so I didn't consider them as critical but maybe I'm missing something.

glennmatthews: I guess it comes back the question of what is the purpose of linting our YAML files at all. :-) If it's to ensure that they're stylistically consistent and avoid basic errors, it should be applied to everything, IMO.

glennmatthews (approved):

PR #95, opened 2021-09-16
Add extra debug information to extract email source

chadell:
Add `emailmessage` information when not able to extract the sender email


glennmatthews (approved):
Code change looks good - what do we need to fix in the `setup-environment` script so that it's working properly? @ubajze ?

chadell:
> Code change looks good - what do we need to fix in the `setup-environment` script so that it's working properly? @ubajze ?

Yes, @ubajze  is working on this in a different branch

PR #100, opened 2021-09-16
Improve traceback formatting in case of exception during process_raw_notification

glennmatthews:
Fix one missing case where an exception traceback wasn't being nicely formatted in Markdown.


chadell (approved):
LGTM

PR #104, opened 2021-09-17
Improve `RawNotification` creation, validating Integrity 

chadell:
Now, we take care as soon as an `Integrity` issue happens, and don't postpone it until the end of the execution as done with the `get_or_create` approach.
Other small changes:
* The Email Notification comes with the `Date` that is the one used as the `Date` for the `RawNotification` being able to understand if this is a duplicated notification in creation time, not waiting for the `Stamp` that will come from the `ParsedNotification`
* Using the `get` and `ObjectDoesNotExist` paradigm to retrieve elements. I believe it's easier to read.

### DIFF
--- a/nautobot_circuit_maintenance/handle_notifications/handler.py
+++ b/nautobot_circuit_maintenance/handle_notifications/handler.py
@@ -3,7 +3,10 @@ import datetime
 import traceback
 from typing import Optional, List
 import uuid
+from dateutil import parser
 from django.conf import settings
+from django.core.exceptions import ObjectDoesNotExist
+from django.db import IntegrityError
 from circuit_maintenance_parser import ProviderError, init_provider, NotificationData, Maintenance
 from nautobot.circuits.models import Circuit, Provider
 from nautobot.extras.jobs import Job, BooleanVar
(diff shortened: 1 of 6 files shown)


nautobot_circuit_maintenance/handle_notifications/handler.py
```diff
@@ -200,21 +202,22 @@ def create_raw_notification(logger: Job, notification: MaintenanceNotification,
...
             source=NotificationSource.objects.filter(name=notification.source).last(),
+            date=parser.parse(notification.date),
         )
-    except Exception as exc:
-        logger.log_warning(message=f"Raw notification '{notification.subject}' not created because {str(exc)}")
-        return None
-
-    if not created:
+        raw_entry.save()
+    except IntegrityError as exc:
```
glennmatthews, line 214: Doesn't `IntegrityError` generally leave the DB in a bad state? Are we able to safely continue processing additional notifications in the case where this happens? In the previous (existing) version of the code, I saw things like this:

image

chadell: My idea capturing the `IntegrityError` was to deal with it and do not touch the database with creating references (for instance with the `CircuitMaintenance`), as it was happening before with the `get_or_create` that was evaluated into the DB later in the process (and the other entities where also "created"). 
Now, as soon as we detect the duplicated entry in the DB, we stop processing and do not create any reference, so we avoid bad state.
If this approach looks flaky, another alternative is to check for an instance with the `unique_together` elements before creating. I thought this was "simpler", but open to suggestions.
I was able to reproduce the issue you are mentioning in my dev environment, and this  fixed the issue for me and also for the new test case added.

glennmatthews: OK, if you have been able to reproduce the issue and verify that this fixes it, I'm OK with this approach. Thanks!

glennmatthews (approved):

Issue #107, opened 2021-09-17
"An error occurred in the current transaction" when encountering duplicate (?) notification


glennmatthews:
### Environment

- Python version: 
- Nautobot version: 1.1.2
- nautobot-circuit-maintenance version: 0.2.3

### Expected Behavior

Circuit maintenance job to run to completion and successfully process notifications, including those that have already been processed.

### Observed Behavior

![image](https://user-images.githubusercontent.com/5603551/133852906-30534340-231f-4670-8d1c-e907994a0636.png)

### Steps to Reproduce

I'm not sure. Even after deleting all pre-existing RawNotification records, I'm encountering this issue, but it does *not* appear that I actually have two duplicate emails (same exact datestamp, subject, and provider) in my inbox, so I don't understand how we're even triggering this condition.

I do have multiple emails from the same provider with the same subject line, but they have different `Date:` stamps on them.

glennmatthews:
Also worth noting that when this condition is encountered, it appears to cause the database to rollback the entire job's changes - when the job completes, there are no new database entries present, even for those notifications that were processed before the error was triggered.

chadell:
I have been able to replicate in my DEV environment, so hopefully it would be solved sooner than later.

chadell:
Test to reproduce it: 60dbb8833538dde11790f1c728427ceaeff577a1

glennmatthews:
Fixed by #108. Thanks!

PR #108, opened 2021-09-18
Verify a RawNotification exists before creating

chadell:
Address #107 

Checking for `IntegrityError` was not good enough to avoid hitting, sporadically, the `atomic block`.
Now the approach uses the `ObjectDoesNotExist` before creating, so not hitting this issue.

### DIFF
--- a/nautobot_circuit_maintenance/handle_notifications/handler.py
+++ b/nautobot_circuit_maintenance/handle_notifications/handler.py
@@ -6,7 +6,6 @@ import uuid
 from dateutil import parser
 from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
-from django.db import IntegrityError
 from circuit_maintenance_parser import ProviderError, init_provider, NotificationData, Maintenance
 from nautobot.circuits.models import Circuit, Provider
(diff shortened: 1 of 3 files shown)


nautobot_circuit_maintenance/handle_notifications/handler.py
```diff
@@ -200,27 +199,31 @@ def create_raw_notification(logger: Job, notification: MaintenanceNotification,
...
+                provider=provider,
+                raw=notification.raw_payload,
+                sender=notification.sender,
+                source=NotificationSource.objects.filter(name=notification.source).last(),
+                date=parser.parse(notification.date),
+            )
+            raw_entry.save()
+            logger.log_success(raw_entry, message="Raw notification created.")
+        except Exception as exc:
+            logger.log_warning(message=f"Raw notification '{notification.subject}' not created because {str(exc)}")
```
chadell, line 225: maybe this should be promoted to `log_failure`? 🤔

carbonarok: This sounds more like a failure to me. What was the reason for using `log_warning`?

carbonarok (approved):
LGTM

Issue #98, opened 2021-09-16
Add email filtering capability when using `source_header`


chadell:
### Proposed Functionality

When using `source_header` different from `From` it is not possible to filter emails, so you get emails that maybe are completely unrelated to the plugin, adding some noise.
Maybe, adding an extra configuration option, such as `source_header_from_emails` that can contain a list of the "senders", this could limit all these issues that are unrelated.

### Use Case

Limit the noise from emails that are not related to the plugin work.

chadell:
Addressed in #109

PR #109, opened 2021-09-20
Add 'source_header_limit_emails' to filter from Gmail API when not using 'From'

chadell:
When using Gmail API, and the `X-Original-Sender` as `source_header`, we are not able to filter emails (#98).
This PR adds a new configuration attribute, for Gmail API configuration, that can be used to limit the notifications processed sent from a mailing list or alias, so omitting emails that has nothing to do with Circuit Maintenance notifications.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,7 @@ There are 2 extra required attributes:
 There are also two optional attributes:
 
 - `source_header`: Specify a particular email header to use to identify the source of a particular notification and assign it to the appropriate provider. If unset, `From` will be used, but if your emails are not received directly from the provider but instead pass through a mailing list or alias, you might need to set this to a different value such as `X-Original-Sender` instead.
+- `source_header_limit_emails`: Restrict the emails used as "From", maybe a email from a mailing list or alias, in order to limit the emails that are processed.
 - `extra_scopes`: Specify a list of additional Google OAuth2 scopes to request access to in addition to GMail API access.
 
 ```py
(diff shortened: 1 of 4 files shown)


carbonarok (approved):
LGTM

README.md
```diff
@@ -121,6 +121,7 @@ There are 2 extra required attributes:
 There are also two optional attributes:
 
 - `source_header`: Specify a particular email header to use to identify the source of a particular notification and assign it to the appropriate provider. If unset, `From` will be used, but if your emails are not received directly from the provider but instead pass through a mailing list or alias, you might need to set this to a different value such as `X-Original-Sender` instead.
+- `source_header_limit_emails`: Restrict the emails used as "From", maybe a email from a mailing list or alias, in order to limit the emails that are processed.
```
glennmatthews, line 124: This doesn't have anything to do with headers - perhaps "limit_to_emails_from" would be a more meaningful name?

glennmatthews (commented):
Anything that can/should be added for unit testing of this feature?

README.md
chadell: I wanted to make it clear that is related to the `source_header`, because the default way to limit emails is by the Providers custom field. 
Gonna try to get some better naming and also some test, I just created this PR quickly to get some feedback on the overall approach

README.md
```diff
@@ -121,6 +121,7 @@ There are 2 extra required attributes:
 There are also two optional attributes:
```
glennmatthews, line 121:
```suggestion
There are also three optional attributes:
```

README.md
```diff
@@ -121,6 +121,7 @@ There are 2 extra required attributes:
...
 
 - `source_header`: Specify a particular email header to use to identify the source of a particular notification and assign it to the appropriate provider. If unset, `From` will be used, but if your emails are not received directly from the provider but instead pass through a mailing list or alias, you might need to set this to a different value such as `X-Original-Sender` instead.
+- `limit_emails_without_header_from`: Restrict the emails used as "From", maybe a email from a mailing list or alias, in order to limit the emails that are processed.
```
glennmatthews, line 124: This is still a non-obvious name to me, but I won't hold up the review about it. :-) Might be good though to add some more explanatory context here, similar to what you have in the CHANGELOG.md above?

glennmatthews (approved):

PR #118, opened 2021-09-22
Avoid creating naive datetime for start_time/end_time

glennmatthews:
When running the job, I noticed that Django complains about `start_time` and `end_time` being naive datetime objects rather than timezone aware:

```
django/db/models/fields/__init__.py:1370: RuntimeWarning: DateTimeField CircuitMaintenance.start_time received a naive datetime (2021-09-15 19:00:00) while time zone support is active.
```

This should fix that issue by ensuring that we explicitly recognize the `parser_maintenance` timestamps as UTC time.


chadell (approved):
Good catch! LGTM

Issue #112, opened 2021-09-20
In CircuitMaintenance detail view, Notification "Date" should not be created/updated timestamp


glennmatthews:
### Environment

- Python version: 
- Nautobot version: 
- nautobot-circuit-maintenance version: 0.2.4

### Expected Behavior

In the CircuitMaintenance detail view, raw notifications associated with a CircuitMaintenance should be listed based on their notification timestamp.

### Observed Behavior

Raw notifications are listed based on their created/updated timestamp, not the notification timestamp.

![image](https://user-images.githubusercontent.com/5603551/134024503-b5505c7e-0f01-4f28-9004-d99a6a1e35ec.png)

Note that all four notifications show times of 9/20 at 2:48 or 2:49 PM as that's the time that the parser job was run and processed these notifications. Clicking through to each individual notification is necessary in order to show its true notification timestamp:

![image](https://user-images.githubusercontent.com/5603551/134024697-c3c9e4b9-a72e-4055-b181-1fa6eaaea9b4.png)

### Steps to Reproduce

1. Run circuit maintenances job with data available in GMail
2. View details of any given CircuitMaintenance record.
3.

chadell:
Addressed in #113

PR #123, opened 2021-09-30
Update CircuitMaintenance only newer notifications

chadell:
Addresses issue #112 

In order to parse notification in **any** order but keep the latest one as the final state, before updating an existing `CircuitMaintenance` we take into account the notification `stamp`, so only the newer ones are taken into account in the final state, but all the parsed notifications are listed in the final `CircuitMaintenance` ordering them via this stamp (HTML template updated)

In the UI test, I had to notifications for the same Maintenance, and now (in debug mode), we see that we are hitting the "Not update CircuitMaintenance...", but the notification is created the same way, and in the rendering is showed in the proper order.

image

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Fix
 
 - #113: In `CircuitMaintenace` detail view, when the parsed notifications are listed, instead of using the date of the parsing action, we show the date when the original notification was sent.
+- #123: Unordered notifications made the final `CircuitMaintenance` state to not being the latest but the one from the last notification parsed. Now, only parsed notifications, newer than the latest one used to update the `CircuitMaintenace` are taken into account, but all of them are linked to it, to have a complete report of them, and be able to render all of them in descendant order in the detail view.
 
 ## v0.2.4 - 2021-09-20
 
--- a/nautobot_circuit_maintenance/handle_notifications/handler.py
+++ b/nautobot_circuit_maintenance/handle_notifications/handler.py
@@ -164,6 +164,18 @@ def create_or_update_circuit_maintenance(
     maintenance_id = f"{raw_entry.provider.slug}-{parser_maintenance.maintenance_id}"
     try:
         circuit_maintenance_entry = CircuitMaintenance.objects.get(name=maintenance_id)
+        last_parsed_notification = (
+            circuit_maintenance_entry.parsednotification_set.order_by("raw_notification__date").reverse().last()
+        )
+
+        # If the notification is older than the latest one used to update the CircuitMaintenance, we skip updating it
+        if last_parsed_notification and last_parsed_notification.date > datetime.datetime.fromtimestamp(
+            parser_maintenance.stamp, tz=datetime.timezone.utc
+        ):
+            if logger.debug:
+                logger.log_debug(f"Not updating CircuitMaintenance {maintenance_id} because the notification is older.")
+            return circuit_maintenance_entry
+
         update_circuit_maintenance(logger, circuit_maintenance_entry, maintenance_id, parser_maintenance, provider)
     except ObjectDoesNotExist:
         circuit_maintenance_entry = create_circuit_maintenance(logger, maintenance_id, parser_maintenance, provider)
@@ -261,7 +273,7 @@ def process_raw_notification(logger: Job, notification: MaintenanceNotification)
             circuit_maintenance_entry = create_or_update_circuit_maintenance(
                 logger, parser_maintenance, raw_entry, provider
             )
-            # Update raw notification as properly parsed and with the stamp time
+            # Update raw notification as properly parsed
             raw_entry.parsed = True
             raw_entry.save()
 
--- a/nautobot_circuit_maintenance/handle_notifications/handler.py
+++ b/nautobot_circuit_maintenance/handle_notifications/handler.py
@@ -173,11 +173,10 @@ def create_or_update_circuit_maintenance(
             parser_maintenance.stamp, tz=datetime.timezone.utc
         )
         if last_parsed_notification and last_parsed_notification.date > parser_maintenance_datetime:
-            if logger.debug:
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ### Fix
 
 - #113: In `CircuitMaintenace` detail view, when the parsed notifications are listed, instead of using the date of the parsing action, we show the date when the original notification was sent.
+- #123: Unordered notifications made the final `CircuitMaintenance` state to not being the latest but the one from the last notification parsed. Now, only parsed notifications, newer than the latest one used to update the `CircuitMaintenace` are taken into account, but all of them are linked to it, to have a complete report of them, and be able to render all of them in descendant order in the detail view.
```
glennmatthews, line 12:
```suggestion
- #123: Out-of-sequence notifications could make the final `CircuitMaintenance` state not reflect the latest notification state. Now, while out-of-sequence notifications will still be processed and linked to the `CircuitMaintenance` for context, they will not result in a change of its overall state.
```

nautobot_circuit_maintenance/handle_notifications/handler.py
```diff
@@ -164,6 +164,18 @@ def create_or_update_circuit_maintenance(
     maintenance_id = f"{raw_entry.provider.slug}-{parser_maintenance.maintenance_id}"
     try:
         circuit_maintenance_entry = CircuitMaintenance.objects.get(name=maintenance_id)
+        last_parsed_notification = (
+            circuit_maintenance_entry.parsednotification_set.order_by("raw_notification__date").reverse().last()
```
glennmatthews, line 168: I see that both `ParsedNotification` and `RawNotification` have a `date` field - for my education, what's the difference between these two fields' expected values? Does sorting the `ParsedNotification` based on the associated `raw_notification__date` result in a different expected ordering than sorting by its own `date` field?

nautobot_circuit_maintenance/handle_notifications/handler.py
```diff
@@ -164,6 +164,18 @@ def create_or_update_circuit_maintenance(
...
+        )
+
+        # If the notification is older than the latest one used to update the CircuitMaintenance, we skip updating it
+        if last_parsed_notification and last_parsed_notification.date > datetime.datetime.fromtimestamp(
+            parser_maintenance.stamp, tz=datetime.timezone.utc
+        ):
+            if logger.debug:
+                logger.log_debug(f"Not updating CircuitMaintenance {maintenance_id} because the notification is older.")
```
glennmatthews, line 176: Might be good to add more specifics to the log message, something like:

```suggestion
        parser_maintenance_datetime = datetime.datetime.fromtimestamp(
            parser_maintenance.stamp, tz=datetime.timezone.utc
        )
        if last_parsed_notification and last_parsed_notification.date > parser_maintenance_datetime:
            if logger.debug:
                logger.log_debug(
                    f"Not updating CircuitMaintenance {maintenance_id} because the notification is from "
                    f"{parser_maintenance_datetime}, older than the most recent notification from {last_parsed_notification.date}."
                )
```

nautobot_circuit_maintenance/views.py
```diff
@@ -34,7 +34,9 @@ def get_extra_context(self, request, instance):
         """Extend content of detailed view for Circuit Maintenance."""
         maintenance_note = models.Note.objects.filter(maintenance=instance)
         circuits = models.CircuitImpact.objects.filter(maintenance=instance)
-        parsednotification = models.ParsedNotification.objects.filter(maintenance=instance)
+        parsednotification = models.ParsedNotification.objects.filter(maintenance=instance).order_by(
+            "raw_notification__date"
+        )
```
glennmatthews, line 39: Same question here - `ParsedNotification` is natively ordered by its own `date` field, what's the intent behind ordering by `raw_notification__date` instead?

glennmatthews (approved):
A few questions and nitpicks, but I like this overall - nice job solving the underlying issue instead of the surface symptom. :-)

nautobot_circuit_maintenance/views.py
glennmatthews: Actually, looking at the screenshot you provided, would it make sense to reverse the ordering so that the newest notification appears at the top of the table?

nautobot_circuit_maintenance/handle_notifications/handler.py
chadell: The ParsedNotification.date is the date when the notification was parsed, the RawNotification.date is the date when the notification was sent.

if it's too confusing we may change the meaning...

glennmatthews: That is a bit confusing - it would be good at least to clearly document the meaning of these fields in `models.py` and wherever they're being referenced, like here.

nautobot_circuit_maintenance/handle_notifications/handler.py
```diff
@@ -164,6 +164,22 @@ def create_or_update_circuit_maintenance(
...
+        last_parsed_notification = (
+            circuit_maintenance_entry.parsednotification_set.order_by("raw_notification__date").reverse().last()
+        )
+
+        # If the notification is older than the latest one used to update the CircuitMaintenance, we skip updating it
+        parser_maintenance_datetime = datetime.datetime.fromtimestamp(
+            parser_maintenance.stamp, tz=datetime.timezone.utc
+        )
+        if last_parsed_notification and last_parsed_notification.date > parser_maintenance_datetime:
+            if logger.debug:
```
glennmatthews, line 176: I think I'd lean towards always logging this message, rather than gating it on the `debug` flag.

glennmatthews (dismissed):
Code looks good - would like to have some comments and/or docstrings explaining the difference between ParsedNotification.date and RawNotification.date and why we're using the latter in this logic.

glennmatthews (approved):

Issue #116, opened 2021-09-22
Increase default `rawnotification` `raw` size to capture at lease email headers


chadell:
### Environment

- Python version: 3.x
- Nautobot version: 1.1.2
- nautobot-circuit-maintenance version: 0.2.4

### Expected Behavior

We should be able to get the basic info from an original email when stored in the `rawnotification.raw` field. This data is truncated by default to 500 characters 

### Observed Behavior

This truncation means missing basic email headers information, so we should look for a value (8192?) that consistently give enough information in the headers to locate the notification in the email source box.

### Steps to Reproduce

1.
2.
3.

glennmatthews:
In many cases neither the end user nor the parser care that much about the details of the email's headers - it would be more useful to specifically capture the parts of the email that are actually used by the parser, i.e. the subject line, sender address, and most importantly the email's actual body content.

chadell:
Changed in #127

PR #127, opened 2021-10-04
Increase RawNotifications raw size and used defaults

chadell:
Addresses #116 
* Use Plugin default settings
* Increase size of raw notification stored: all the headers are included.

### DIFF
--- a/nautobot_circuit_maintenance/__init__.py
+++ b/nautobot_circuit_maintenance/__init__.py
@@ -84,7 +84,12 @@ class CircuitMaintenanceConfig(PluginConfig):
     min_version = "1.0.0-beta.4"
     max_version = "1.999"
     required_settings = []
-    default_settings = {}
+    default_settings = {
+        "raw_notifications": {
+            "initial_days_since": 365,
+            "raw_notification_size": 8192,
+        },
+    }
     caching_config = {}
 
     def ready(self):
--- a/nautobot_circuit_maintenance/__init__.py
+++ b/nautobot_circuit_maintenance/__init__.py
@@ -85,10 +85,8 @@ class CircuitMaintenanceConfig(PluginConfig):
     max_version = "1.999"
     required_settings = []
     default_settings = {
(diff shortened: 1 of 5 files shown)


nautobot_circuit_maintenance/__init__.py
```diff
@@ -84,7 +84,12 @@ class CircuitMaintenanceConfig(PluginConfig):
     min_version = "1.0.0-beta.4"
     max_version = "1.999"
     required_settings = []
-    default_settings = {}
+    default_settings = {
+        "raw_notifications": {
+            "initial_days_since": 365,
+            "raw_notification_size": 8192,
+        },
+    }
```
glennmatthews, line 92: Does this make it so that if I change `raw_notification_size` in my `nautobot_config.py`, I have to also define a value for `initial_days_since` to avoid an error? It seems like setting:

```python
PLUGINS_CONFIG={
    "nautobot_circuit_maintenance": {
        "raw_notifications": {"raw_notification_size": 16384}
    }
}
```

would result in `settings["PLUGINS_CONFIG"]["nautobot_circuit_maintenance"]["raw_notifications"]["initial_days_since"]` becoming undefined.

glennmatthews (commented):
As I stated in #116, to me the headers are the least intereresting part of an email notification. I'd much rather see this plugin capturing the full email body (or even just the first 4k or 8k bytes, if it's a big email) in the RawNotification.

nautobot_circuit_maintenance/__init__.py
chadell: true

chadell:
> As I stated in #116, to me the headers are the least intereresting part of an email notification. I'd much rather see this plugin capturing the full email body (or even just the first 4k or 8k bytes, if it's a big email) in the RawNotification.

Yes, defaulting to 8K it's a good size to get good email representation without overwhelming the DB

glennmatthews (approved):

Issue #126, opened 2021-10-02
The `get_since_reference` provides is not consistent enough


chadell:
### Environment

- Python version: 
- Nautobot version: 1.1.3
- nautobot-circuit-maintenance version: 0.2.5

### Expected Behavior

The reference time to get notifications should be the last one that we got raw_notifications. Because now the `date` is matching the `STAMP` or the `DATE` of the email, this causes that a wrong STAMP, maybe in the future, will avoid retrieving notifications

### Steps to Reproduce

1. Fake a ICS notificaitons with a STAMP in the future
2. Run the job and the `since` will be in the future
3.

chadell:
fixed by #128

PR #128, opened 2021-10-04
`RawNotification` contains a `date` and `stamp`

chadell:
In order to solve issue #126 , `RawNotification` will contain `stamp`: the original date of the `RawNotification` was sent, that can be in the past as it's independent of the processing time.

This PR also includes a python script to copy old `date` to `stamp` as `date` was referencing to the `stamp` meaning. 
`stamp` is mandatory and it's not autogenerated, also it is validated that the stamp doesn't come from the future (ping @tim-fiola )

It also removes the `date` from `RawNotification`, `ParsedNotification` and `Note` and using `created` and `last_updated`


glennmatthews (commented):
With this change, we now have RawNotification.date, RawNotification.stamp, ParsedNotification.date, and Note.date to keep track of. Additionally, via Nautobot core, we also have `created` (date) and `last_updated` (datetime) fields on CircuitMaintenance, CircuitImpact, Note, NotificationSource, RawNotification, and ParsedNotification. That's a lot of different datestamps and timestamps. 

Is there anything we can do to *reduce* the number of distinct such fields? For example, can we get rid of `ParsedNotification.date` now, isn't it redundant with `RawNotification.date`?

nautobot_circuit_maintenance/models.py
```diff
@@ -267,6 +269,13 @@ def to_csv(self):
         """Return fields for bulk view."""
         return (self.subject, self.provider, self.sender, self.source, self.raw, self.date, self.parsed)
 
+    def clean(self):
+        """Add validation when creating a RawNotification."""
+        super().clean()
+
+        if self.date < self.stamp:
```
glennmatthews, line 276: Shouldn't this be something like `if self.stamp > datetime.now(timezone.utc)` instead? `self.date` can be changed to any arbitrary date.

Also I hope that Nautobot is always being run on an NTP-managed system with this change, I would hate to see it start throwing errors just because of a few seconds of clock drift resulting in it receiving perfectly valid notifications "from the future"...!

nautobot_circuit_maintenance/templates/nautobot_circuit_maintenance/rawnotification.html
```diff
@@ -59,6 +59,10 @@ <h1>{{ object.subject }}</h1>
             {% else %} {{ object.parsed }} {% endif %}
           </td>
         </tr>
+        <tr>
+          <td>Stamp</td>
+          <td>{{ object.stamp }}</td>
+        </tr>
         <tr>
           <td>Date</td>
```
glennmatthews, line 67: Can we make these more descriptive to the user? Perhaps "Notification Timestamp" and "Processing Date" or something like that would be accurate and more descriptive?

nautobot_circuit_maintenance/models.py
glennmatthews: Also it doesn't look like you have test coverage for this check. :-)

chadell:
> Is there anything we can do to _reduce_ the number of distinct such fields? For example, can we get rid of `ParsedNotification.date` now, isn't it redundant with `RawNotification.date`?

this is right, I will leverage all the "date" on the default timestamps, apart from the `stamp` one

nautobot_circuit_maintenance/handle_notifications/handler.py
```diff
@@ -165,20 +165,19 @@ def create_or_update_circuit_maintenance(
     try:
         circuit_maintenance_entry = CircuitMaintenance.objects.get(name=maintenance_id)
         # Using the RawNotification.date as the reference to sort because it's the one that takes into account the
```
glennmatthews, line 167:
```suggestion
        # Using the RawNotification.stamp as the reference to sort because it's the one that takes into account the
```

nautobot_circuit_maintenance/handle_notifications/handler.py
```diff
@@ -305,10 +304,10 @@ def process_raw_notification(logger: Job, notification: MaintenanceNotification)
 
 def get_since_reference(logger: Job) -> int:
     """Get the timestamp from the latest processed RawNotification or a reference from config `initial_days_since`."""
-    # Latest retrieved notification will limit the scope of notifications to retrieve
+    # Latest processed RawNotification will limit the scope of notifications to retrieve
     last_raw_notification = RawNotification.objects.last()
     if last_raw_notification:
-        since_reference = last_raw_notification.date.timestamp()
+        since_reference = last_raw_notification.created.timestamp()
```
glennmatthews, line 310: Just a note that `created` is a date, not a datetime. So there's some loss in granularity here - want to make sure we don't end up missing notifications as a result.

nautobot_circuit_maintenance/models.py
```diff
@@ -265,7 +264,14 @@ def get_absolute_url(self):
...
         """Return fields for bulk view."""
-        return (self.subject, self.provider, self.sender, self.source, self.raw, self.date, self.parsed)
+        return (self.subject, self.provider, self.sender, self.source, self.raw, self.stamp, self.parsed)
+
+    def clean(self):
+        """Add validation when creating a RawNotification."""
+        super().clean()
+
+        if self.stamp > datetime.utcnow():
+            raise ValidationError(f"Stamp time {self.stamp} is not consistent, it's in the future.")
```
glennmatthews, line 274: The more I think about this, the less I like it. If there's clock drift or something I don't want us to reject otherwise valid notifications. Maybe change this to a `logger.warning()` if we need it at all?

nautobot_circuit_maintenance/handle_notifications/handler.py
chadell: The `SINCE` filter for the email search is also a `date` not a `datetime`, so we are good.

nautobot_circuit_maintenance/models.py
chadell: Ok, I will lower this to just a log message. Not a big deal at all, because the important reference now is the `created`: `since_reference = last_raw_notification.created.timestamp()`

chadell:
@glennmatthews , could you check this PR  again ?

nautobot_circuit_maintenance/tests/test_api.py
```diff
@@ -32,8 +32,8 @@ def setUpTestData(cls):
         cls.create_data = [
             {
                 "name": "UT-TEST1",
-                "start_time": now() + datetime.timedelta(days=5),
-                "end_time": now() + datetime.timedelta(days=6),
+                "start_time": utcnow() + datetime.timedelta(days=5),
+                "end_time": utcnow() + datetime.timedelta(days=6),
```
glennmatthews, line 36: The [Python datetime docs](https://docs.python.org/3/library/datetime.html#datetime.datetime.now) recommend using `now(timezone.utc)` instead of `utcnow()` as the former creates a timezone-aware UTC time whereas the latter creates a naive (timezone-unaware) time. We're doing our best to use aware timestamps in the actual production code for this plugin, is there a a reason not to do so in the tests as well?

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## v0.2.6 -
+
+## Changed
+
+- #127: Increase default `raw_notification_size` from 1000 to 8192, and also update this and `raw_notification_initial_days_since` definition in `PLUGINS_CONFIG`.
```
glennmatthews, line 7: Given this is a backwards-incompatible change in the plugin configuration, should we treat this commit as beginning a 0.3.0 release rather than a 0.2.x patch release?

nautobot_circuit_maintenance/models.py
```diff
@@ -262,7 +264,14 @@ def get_absolute_url(self):
...
     def to_csv(self):
         """Return fields for bulk view."""
-        return (self.subject, self.provider, self.sender, self.source, self.raw, self.date, self.parsed)
+        return (self.subject, self.provider, self.sender, self.source, self.raw, self.stamp, self.parsed)
+
+    def clean(self):
+        """Add validation when creating a RawNotification."""
+        super().clean()
+
+        if self.stamp > datetime.utcnow():
```
glennmatthews, line 273: Should be using `datetime.now(timezone.utc)` to remain timezone-aware, I think.

nautobot_circuit_maintenance/templates/nautobot_circuit_maintenance/note.html
```diff
@@ -46,8 +46,8 @@ <h1>{{ object.title }}</h1>
             <td>{{ object.comment }}</td>
         </tr>
         <tr>
-            <td>Date</td>
-            <td>{{ object.date }}</td>
+            <td>Last Updated</td>
```
glennmatthews, line 49: Not blocking for this PR, but it did make me notice that although we're using change-logged base classes for all of these models, our page templates don't provide a tab for viewing the changelog. (You can see that the changelog exists from the Nautobot home page summary, though.) We should fix that.

(When/if we move to requiring Nautobot 1.2, this will be made a lot easier by the implementation of https://github.com/nautobot/nautobot/pull/737)

CHANGELOG.md
chadell: good point

nautobot_circuit_maintenance/tests/test_api.py
chadell: will do too

nautobot_circuit_maintenance/templates/nautobot_circuit_maintenance/note.html
chadell: @glennmatthews , do you mean about [this](https://github.com/nautobot/nautobot-plugin-circuit-maintenance/blob/develop/nautobot_circuit_maintenance/templates/nautobot_circuit_maintenance/circuitmaintenance.html#L39-L43)?
is it only supported for `CircuitMaintenance` and `CircuitImpact`. I would add an issue to update the other in a different PR

glennmatthews: Yeah, that part of the template. I guess the main one I noticed it missing from was `Note`. I'll open an issue.

nautobot_circuit_maintenance/handle_notifications/handler.py
```diff
@@ -163,21 +163,20 @@ def create_or_update_circuit_maintenance(
...
         # If the notification is older than the latest one used to update the CircuitMaintenance, we skip updating it
         parser_maintenance_datetime = datetime.datetime.fromtimestamp(
             parser_maintenance.stamp, tz=datetime.timezone.utc
         )
-        if last_parsed_notification and last_parsed_notification.date > parser_maintenance_datetime:
+        if last_parsed_notification and last_parsed_notification.last_updated > parser_maintenance_datetime:
             logger.log_debug(
                 f"Not updating CircuitMaintenance {maintenance_id} because the notification is from "
-                f"{parser_maintenance_datetime}, older than the most recent notification from {last_parsed_notification.date}."
+                f"{parser_maintenance_datetime}, older than the most recent notification from {last_parsed_notification.created}."
```
glennmatthews, line 179: Should this be using `last_updated` rather than `created` for consistency with the actual logic above?

glennmatthews (dismissed):

nautobot_circuit_maintenance/handle_notifications/handler.py
chadell: totally

PR #129, opened 2021-10-05
Improve response time of app metrics endpoint

dgarros:
After enabling the app metrics endpoint in a production environment we noticed that the response time was between 3 and 5 seconds for few hundred circuits. 

After some investigation, I was able to reduce the response time significantly by adding some select_related tables to the main query.
In my test enviroments, it improved the response time by 100x

### DIFF
--- a/nautobot_circuit_maintenance/metrics_app.py
+++ b/nautobot_circuit_maintenance/metrics_app.py
@@ -60,7 +60,7 @@ def metric_circuit_operational():
         .prefetch_related("circuit")
     )
 
-    for termination in CircuitTermination.objects.all().prefetch_related("circuit"):
+    for termination in CircuitTermination.objects.all().select_related("circuit", "circuit__provider", "circuit__type", "site"):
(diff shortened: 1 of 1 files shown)


chadell (dismissed):
LGTM, thanks @dgarros , makes sense to optimize for the `DEFAULT_LABELS`, fetching the 4 attributes that it provides by default.
Maybe, to make this evident, I would add a note in the code to "link" the tables selected to it, so in the future we change one, we update the other.

glennmatthews (approved):

dgarros:
> LGTM, thanks @dgarros , makes sense to optimize for the `DEFAULT_LABELS`, fetching the 4 attributes that it provides by default. Maybe, to make this evident, I would add a note in the code to "link" the tables selected to it, so in the future we change one, we update the other.

Good point I added a comment

chadell (approved):
LGTM

glennmatthews (approved):

PR #135, opened 2021-10-14
Use Graphql extra feature

chadell:


chadell:
I verified that using `extra_features` was colliding with the craft types:
```
        nautobot_1       | 05:11:28.367 WARNING nautobot.graphql.schema schema.py                 generate_query_mixin() :
nautobot_1       |   Unable to load schema type for the model "nautobot_circuit_maintenance.circuitmaintenance" as there is already another type registered under this name. If you are seeing this message during plugin development, check to make sure that you aren't using @extras_features("graphql") on the same model you're also defining a custom GraphQL type for.
nautobot_1       | 05:11:28.368 WARNING nautobot.graphql.schema schema.py                 generate_query_mixin() :
nautobot_1       |   Unable to load schema type for the model "nautobot_circuit_maintenance.circuitimpact" as there is already another type registered under this name. If you are seeing this message during plugin development, check to make sure that you aren't using @extras_features("graphql") on the same model you're also defining a custom GraphQL type for.
```
so removing non necessary code

chadell:
@glennmatthews I've added testing for GraphQL. 
As it was already approved (and only cleanup and tests have been added), I would move it forward when tests pass.

PR #151, opened 2021-11-08
Add support for labeling of Gmail messages based on outcome

### LINKED ISSUE #120: GMail integration: optionally tag/categorize emails when successfully parsed.
### Environment

- Nautobot version: 
- nautobot-circuit-maintenance version: 

### Proposed Functionality

In the GMail integration, it would be very useful to tag processed emails with one or more categories after processing them, e.g. "parsed" and "parse_failed" perhaps.

### Use Case

This makes it easier for the developer or user to filter the service inbox to see which messages were processed, skipped, and/or failed parsing for troubleshooting and data accuracy checking.

glennmatthews:
Fixes #120. To do this I had to change some of the data that's passed around during execution of the job so as to ensure that we have access to the original `MaintenanceNotification` object (and its associated `Source` object) wherever we need to make a call back to tag the original email with a new label.

I also added `docs/images/data_flow.png` as a developer reference file because I kept getting lost in the weeds without it. :-) There are a lot of very similarly named classes in this plugin, some of which are DB models and some of which are in-memory Pydantic objects, and it's easy to get mixed up.

Marking this as a draft for the moment because I haven't yet had time to test it against a live Gmail account to verify that the `messages.modify` API call is correct, but I believe the code is otherwise complete.


README.md
```diff
@@ -116,11 +116,18 @@ There are 2 extra required attributes:
...
 
-There are also three optional attributes:
+There are also the following optional attributes:
 
 - `source_header`: Specify a particular email header to use to identify the source of a particular notification and assign it to the appropriate provider. If unset, `From` will be used, but if your emails are not received directly from the provider but instead pass through a mailing list or alias, you might need to set this to a different value such as `X-Original-Sender` instead.
 - `limit_emails_with_not_header_from`: List of emails used to restrict the emails retrieved when NOT using the `source_header` "From" and we can't use the `Provider` original emails to filter.
 - `extra_scopes`: Specify a list of additional Google OAuth2 scopes to request access to in addition to GMail API access.
+- `labels`: Specify a dictionary of message labels and their corresponding Gmail label IDs; used to automatically label messages as they are processed by this plugin for later investigation and troubleshooting. Any labels that are not specified will be skipped. Currently supported labels are:
```
chadell, line 124: should we add a note in the Readme to "how to create and read labels ids"?

nautobot_circuit_maintenance/handle_notifications/sources.py
```diff
@@ -38,7 +39,8 @@
 class MaintenanceNotification(BaseModel):
     """Representation of all the data related to a Maintenance Notification."""
 
-    source: str
+    msg_id: bytes
+    source: "Source"
```
chadell, line 43: here we lose track of the Source name, and add hardcode this string `Source`. Maybe we were not actually using the `source` before, but for which purpose we use this string now?

nautobot_circuit_maintenance/handle_notifications/sources.py
```diff
@@ -626,3 +648,6 @@ def get_notifications(
             )
 
     return received_notifications
+
+
+MaintenanceNotification.update_forward_refs()
```
chadell, line 653: could you remind me why do we need to apply self referencing types? maybe I missed, but I don't see it self referencing

README.md
```diff
@@ -152,7 +167,7 @@ To create a [Service Account](https://support.google.com/a/answer/7378726?hl=en)
 3. Still under **APIs and Services**, in **Credentials**, create a new **Service Account** and save the credentials file generated to be used when configuring Nautobot Sources.
 4. With Admin rights, edit the newly created Service Account and expand the **Show Domain-Wide Delegation** section. Enable Google Workspace Domain-wide Delegation and save the changes. Copy the Client ID shown.
 5. With Super Admin rights, open the [Google Workspace admin console](https://admin.google.com). Navigate to **Security**, **API controls**, and select the **Manage Domain Wide Delegation** at the bottom of the page.
-6. Add a new API client and paste in the Client ID copied earlier. In the **OAuth scopes** field add the scopes `https://www.googleapis.com/auth/gmail.readonly` and `https://mail.google.com/`. Save the new client configuration by clicking _Authorize_.
+6. Add a new API client and paste in the Client ID copied earlier. In the **OAuth scopes** field add the scopes `https://www.googleapis.com/auth/gmail.readonly`, `https://www.googleapis.com/auth/gmail.modify`, and `https://mail.google.com/`. Save the new client configuration by clicking _Authorize_.
```
chadell, line 170: is this `modify` optional? I mean, if you don't want to update the labels you don't have to add it

nautobot_circuit_maintenance/handle_notifications/sources.py
chadell: follow-up question. If we decided to move `Source` before this part of the code, could we have used `source: Source` directly, no? and doing so, could we have omitted the `update_forward_refs()`?

nautobot_circuit_maintenance/handle_notifications/sources.py
```diff
@@ -407,10 +420,14 @@ class GmailAPI(EmailSource):
     service: Optional[Resource] = None
     credentials: Optional[Union[service_account.Credentials, Credentials]] = None
 
-    SCOPES = ["https://www.googleapis.com/auth/gmail.readonly"]
+    SCOPES = [
+        "https://www.googleapis.com/auth/gmail.readonly",
+        "https://www.googleapis.com/auth/gmail.modify",
```
chadell, line 425: what if someone has concerns on autorizing this new scope and doesn't care about this feature? I believe we should leave this open and maybe add it as an extra_scope if explicitly defined via configuration

chadell (commented):
I like the overall idea, my main concern is about making the new scope mandatory to use Gmail integration

nautobot_circuit_maintenance/handle_notifications/sources.py
glennmatthews: That's a fair point... I'll see how much work it'll be to make it optional.

nautobot_circuit_maintenance/handle_notifications/sources.py
glennmatthews: Yes, that would be an alternate approach, assuming there aren't forward references to `MaintenanceNotification` within the `Source` class that would be equally problematic.

glennmatthews: Turned out to be easy enough to move MaintenanceNotification below Source, so I've done so.

glennmatthews:
CI failure is due to GitHub actions moving to Python 3.9.8 causing Black to hit https://github.com/python/typed_ast/issues/169. Longer term fix is to move to a newer version of Black, but in the short term a new release of `typed_ast` is supposedly coming today with a fix for this issue.

Tests all pass locally using a different version of Python.

steve-mavens:
@glennmatthews FWIW I found the migration black 20.8b1 -> 21.10b0 to be painless. Unlike certain previous black version bumps I've suffered, it didn't modify any of my code.

chadell (approved):
LGTM

PR #156, opened 2021-11-12
Use case-insensitive lookups when matching CIDs to Circuits

glennmatthews:
Human data entry being what it is, I've encountered a number of cases where the capitalization of CIDs in notification emails differs from the capitalization of the CIDs in the Nautobot database. This PR allows the plugin to gracefully handle such differences by using case-insensitive comparisons and logic where appropriate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


chadell (approved):
LGTM

PR #159, opened 2021-11-29
Setting up ci for mysql and postgres test

chadell:
* Add Mysql unittest to CI
* Fix plugin usage from tests

Manual tests:
* Tested user interaction using Mysql
* Tested user interaction using Postgres
* Tested data persistency in Postrgres from previous release to the new one after migrations


pyproject.toml
```diff
@@ -30,6 +30,9 @@ google-api-python-client = "^2.9.0"
 google-oauth = "^1.0.0"
 google-auth-httplib2 = "^0.1.0"
 google-auth-oauthlib = "^0.4.4"
+# Added pinned typed-add to overcome issues with black version
+typed-ast = "^1.5.0"
```
glennmatthews, line 34: If this is a Black requirement, it should be a dev dependency not a main dependency.

pyproject.toml
```diff
@@ -42,6 +45,9 @@ pylint-django = "*"
 pydocstyle = "*"
 flake8 = "*"
 parameterized = "*"
+python-dotenv = "^0.19.1"
+mysqlclient = "^2.0.3"
```
glennmatthews, line 49: These should be `"*"` versions, especially `mysqlclient` so that it won't conflict when Nautobot changes its dependency version.

pyproject.toml
```diff
@@ -92,6 +98,8 @@ disable = """,
     too-few-public-methods,
     duplicate-code,
     too-many-ancestors,
+    unspecified-encoding,
+    consider-using-f-string,
```
glennmatthews, line 102: It's a step backwards IMHO for us to add new pylint global disables. Better to fix these or disable them case-by-case if absolutely necessary.

tasks.py
```diff
@@ -20,6 +21,26 @@ def is_truthy(arg):
...
+
+    Returns:
+        list: List of docker-compose files.
+    """
+    files = [
+        "docker-compose.redis.yml",
+        "docker-compose.base.yml",
+        "docker-compose.dev.yml",
+    ]
+    env_vars = dotenv_values(os.path.join(os.path.dirname(__file__), "development/dev.env"))
```
glennmatthews, line 35:
```suggestion
    env_vars = dotenv_values(os.path.join(os.path.dirname(__file__), "development", "dev.env"))
```

pyproject.toml
chadell: sure, I am just trying one to make it fail due mysql, and then I will refine the whole PR

chadell:
good, now we have it where we want it, failing due `mysql`:
`django.db.utils.OperationalError: (1170, "BLOB/TEXT column 'raw' used in key specification without a key length")`
time to fix it 🔧

chadell:
@glennmatthews one of the hottest topics here is how to manage the migrations that need to be update to be applied to Mysql (because of different restrictions).
The option I took was to update the current migrations to make it work: https://github.com/nautobot/nautobot-plugin-circuit-maintenance/pull/159/commits/cd695c279e99ec15746b51c03f46e9e3171aeac3
Looking for your feedback to understand if there is a better way of getting this working

dgarros (approved):
As far as I can tell, looks great. 
I haven't looked too closely at the migrations but I trust it's working for existing deployments

.github/workflows/ci.yml
```diff
@@ -113,13 +113,39 @@ jobs:
...
       - name: "Setup environment"
-        uses: "networktocode/gh-action-setup-poetry-environment@v1"
+        uses: "networktocode/gh-action-setup-poetry-environment@v2"
       - name: "Copy credentials"
         run: "cp development/creds.example.env development/creds.env"
       - name: "Build Container"
         run: "poetry run invoke build"
       - name: "Run Tests"
-        run: "poetry run invoke unittest"
+        run: "poetry run invoke unittest --failfast --keepdb"
```
glennmatthews, line 122: Not sure what benefit `keepdb` provides here?

nautobot_circuit_maintenance/models.py
```diff
@@ -133,7 +134,7 @@ class Note(OrganizationalModel):
     """Model for maintenance notes."""
 
     maintenance = models.ForeignKey(CircuitMaintenance, on_delete=models.CASCADE, default=None)
-    title = models.TextField()
+    title = models.CharField(max_length=500)
```
glennmatthews, line 137: I thought that the issue with Django + MySQL + TextField was only an issue with fields that have `unique=True`? Why does `title` need to be changed from `TextField` to `CharField`, but `comment` (below) does not?

nautobot_circuit_maintenance/models.py
```diff
@@ -22,6 +22,7 @@
 MAX_MAINTENANCE_NAME_LENGTH = 100
 MAX_NOTIFICATION_SENDER_LENGTH = 200
 MAX_NOTIFICATION_SUBJECT_LENGTH = 200
+MAX_NOTIFICATION_TOTAL_LENGTH = 16384
```
glennmatthews, line 25: If we need to impose this limit, then this should be documented in the README as the maximum permissible value for the `raw_notification_size` config setting.

.github/workflows/ci.yml
chadell: nothing, it was left behind

chadell, line 122:
```suggestion
        run: "poetry run invoke unittest --failfast"
```

nautobot_circuit_maintenance/models.py
chadell: It's actually the case: https://github.com/nautobot/nautobot-plugin-circuit-maintenance/blob/develop/nautobot_circuit_maintenance/models.py#L151

glennmatthews: I missed that, thanks!

What do you think about maybe limiting this to 100 or 200 characters? 500 seems excessive for a `title` string, unless you're aware of cases that already exist where we might need that many?

The limit should be another constant in any case, rather than a literal `500` here. Also we should make sure that anywhere in the code where we're creating or updating a Note's `title` has safety code to truncate the string to fit into the DB field if needed, similar to what we have for maintenance names, notification subject lines, etc.

.github/workflows/ci.yml
```diff
@@ -113,13 +113,39 @@ jobs:
...
+      - name: "Check out repository code"
+        uses: "actions/checkout@v2"
+      - name: "Setup environment"
+        uses: "networktocode/gh-action-setup-poetry-environment@v2"
+      - name: "Copy credentials"
+        run: "cp development/creds.example.env development/creds.env"
+      - name: "Build Container"
+        run: "poetry run invoke build"
+      - name: "Run Tests"
+        run: "poetry run invoke unittest --failfast --keepdb"
```
glennmatthews, line 148: Leftover `keepdb` here as well?

README.md
```diff
@@ -335,6 +335,14 @@ Each command can be executed with `invoke <command>`. All commands support the a
   unittest         Run Django unit tests for the plugin.
 ```
 
+`unittest` uses by default the Postgres DB backend, if you want to test Mysql you must use Invoke environment variables:
```
glennmatthews, line 338: This isn't specific to `invoke unittest` - all of the development environment will use postgres by default. As such I think it'd be better to move this addition up to the parent `CLI Helper Commands` section.

README.md
```diff
@@ -335,6 +335,14 @@ Each command can be executed with `invoke <command>`. All commands support the a
...
+
+```bash
+export INVOKE_NAUTOBOT_CIRCUIT_MAINTENANCE_NAUTOBOT_DB_ENGINE=django.db.backends.mysql
+export INVOKE_NAUTOBOT_CIRCUIT_MAINTENANCE_NAUTOBOT_DB_PORT=3306
+export INVOKE_NAUTOBOT_CIRCUIT_MAINTENANCE_NAUTOBOT_DB_USER=root
```
glennmatthews, line 343: Feels a bit weird to me that setting `INVOKE_NAUTOBOT_CIRCUIT_MAINTENANCE_NAUTOBOT_DB_ENGINE` is really just a roundabout way to set `NAUTOBOT_DB_ENGINE` that actually influences the development environment. Why not just instruct the user to update their `creds.env` file directly instead?

nautobot_circuit_maintenance/models.py
chadell: Ok, let's limit it to the 200 characters, and truncate it when necessary.

development/docker-compose.mysql.yml
```diff
@@ -0,0 +1,21 @@
...
+    volumes:
+      - "lcm_mysql_data:/var/lib/mysql"
+    # ports:
+    #   - "3306:3306"
+    healthcheck:
+      test: ["CMD", "mysqladmin", "ping", "-h", "localhost"]
+      timeout: "20s"
+      retries: "10"
+volumes:
+  lcm_mysql_data:  # yamllint disable-line rule:empty-values
```
glennmatthews, line 21: What's the `lcm_` stand for?

.github/workflows/ci.yml
chadell: yes, I am a consistent copy-paster 🤦

development/docker-compose.mysql.yml
chadell: old reference, gonna remove

README.md
chadell: I did this way to change the behaviour in the CI: https://github.com/nautobot/nautobot-plugin-circuit-maintenance/pull/159/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR135
Using your proposal, I would have to define to `creds.example.env`, one for Postgres and the other for Mysql, and copy one or the other.
I see both as valid, so if you prefer the one you are proposing, I can change it

chadell: Something that it's not fully covered with the `creds.env` approach is that we also use this variable to define the [docker-compose files to use](https://github.com/nautobot/nautobot-plugin-circuit-maintenance/pull/159/files#diff-2e746c9eacad6e2c2fdefdc6a665b5ce4607ea384ebf97a729044b79acc778d6R30) (apart of the Nautobot config settings), so you still need to export the `invoke` variables, at least the one for `engine`.

chadell:
@glennmatthews what do you think about the changes in the migrations to make them compatible with Mysql? do you see any potential drawback?

glennmatthews:
I presume you've tested with deploying the updated plugin atop an existing PostgreSQL deployment with existing data? Did everything get preserved correctly?

chadell:
> I presume you've tested with deploying the updated plugin atop an existing PostgreSQL deployment with existing data? Did everything get preserved correctly?

Not yet, I was waiting for your guidance on that.
so, to test it, is it simple as have data from a previous release and run the new/modified migrations on to of it?
does it make sense to test migrations from older versions?

glennmatthews:
> > I presume you've tested with deploying the updated plugin atop an existing PostgreSQL deployment with existing data? Did everything get preserved correctly?
> 
> Not yet, I was waiting for your guidance on that. so, to test it, is it simple as have data from a previous release and run the new/modified migrations on to of it? does it make sense to test migrations from older versions?

Oh gosh. I didn't realize you hadn't tested the migration changes at all yet. :-( I have no idea if the approach you're following here will work or not - definitely need to test with data from a previous release, install the new release, and check whether `nautobot-server showmigrations` and `nautobot-server migrate` work or complain.

chadell:
> > > I presume you've tested with deploying the updated plugin atop an existing PostgreSQL deployment with existing data? Did everything get preserved correctly?
> > 
> > 
> > Not yet, I was waiting for your guidance on that. so, to test it, is it simple as have data from a previous release and run the new/modified migrations on to of it? does it make sense to test migrations from older versions?
> 
> Oh gosh. I didn't realize you hadn't tested the migration changes at all yet. :-( I have no idea if the approach you're following here will work or not - definitely need to test with data from a previous release, install the new release, and check whether `nautobot-server showmigrations` and `nautobot-server migrate` work or complain.

No problem @glennmatthews , I finally done the pending tests and did some minor adjustments, and it looks like everything is working.

Manual tests done:
* Tested user interaction to fetch emails from Gmail Service Account with Postgres and Mysql in last release
* Tested data persistency from previous release to the new one in Postgres

glennmatthews:
What is the Nautobot console output when you start the server for the first time after upgrading from a previous version of the plugin?

chadell:
> What is the Nautobot console output when you start the server for the first time after upgrading from a previous version of the plugin?

Only DB and Nautobot server output:

```
db_1             | 
db_1             | PostgreSQL Database directory appears to contain a database; Skipping initialization
db_1             | 
db_1             | 2021-12-17 14:44:17.841 UTC [1] LOG:  starting PostgreSQL 14.0 on x86_64-pc-linux-musl, compiled by gcc (Alpine 10.3.1_git20210424) 10.3.1 20210424, 64-bit
db_1             | 2021-12-17 14:44:17.841 UTC [1] LOG:  listening on IPv4 address "0.0.0.0", port 5432
db_1             | 2021-12-17 14:44:17.841 UTC [1] LOG:  listening on IPv6 address "::", port 5432
db_1             | 2021-12-17 14:44:17.846 UTC [1] LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"
db_1             | 2021-12-17 14:44:17.853 UTC [22] LOG:  database system was interrupted; last known up at 2021-12-17 14:35:08 UTC
db_1             | 2021-12-17 14:44:18.193 UTC [22] LOG:  database system was not properly shut down; automatic recovery in progress
db_1             | 2021-12-17 14:44:18.196 UTC [22] LOG:  redo starts at 0/16FAD90
db_1             | 2021-12-17 14:44:18.248 UTC [22] LOG:  invalid record length at 0/1BFCEF0: wanted 24, got 0
db_1             | 2021-12-17 14:44:18.248 UTC [22] LOG:  redo done at 0/1BFCEC8 system usage: CPU: user: 0.03 s, system: 0.01 s, elapsed: 0.05 s
db_1             | 2021-12-17 14:44:18.459 UTC [1] LOG:  database system is ready to accept connections
nautobot_1       | Performing database migrations...
nautobot_1       | Operations to perform:
nautobot_1       |   Apply all migrations: admin, auth, circuits, contenttypes, dcim, extras, ipam, nautobot_circuit_maintenance, sessions, social_django, taggit, tenancy, users, virtualization
nautobot_1       | Running migrations:
nautobot_1       |   Applying extras.0012_healthchecktestmodel... OK
nautobot_1       |   Applying extras.0013_default_fallback_value_computedfield... OK
nautobot_1       |   Applying nautobot_circuit_maintenance.0011_mysql... OK
nautobot_1       |   Applying virtualization.0003_vminterface_verbose_name... OK
nautobot_1       | 
nautobot_1       | Generating cable paths...
...
nautobot_1       | Finished.
nautobot_1       | 
nautobot_1       | Collecting static files...
nautobot_1       | 
nautobot_1       | 972 static files copied to '/opt/nautobot/static'.
nautobot_1       | 
nautobot_1       | Removing stale content types...
nautobot_1       | 
nautobot_1       | Removing expired sessions...
nautobot_1       | 
nautobot_1       | System check identified some issues:
nautobot_1       | 
nautobot_1       | WARNINGS:
...
nautobot_1       | Performing system checks...
nautobot_1       | 
nautobot_1       | System check identified no issues (0 silenced).
nautobot_1       | December 17, 2021 - 14:45:18
nautobot_1       | Django version 3.1.13, using settings 'nautobot_config'
nautobot_1       | Starting development server at http://0.0.0.0:8080/
nautobot_1       | Quit the server with CONTROL-C.
nautobot_1       | 14:45:18.579 INFO    nautobot             __init__.py                              setup() :
nautobot_1       |   Nautobot initialized!
```

glennmatthews:
Good to see that it applies the `11_mysql` migration but I'm concerned that the updates made to the earlier migrations aren't getting applied. I suspect that if you check the DB after this migration, for instance, `Note.title` is still a TextField (`varchar`?) rather than a 512-character-max-length CharField, which is different than you'll probably see if you create the DB fresh from scratch with all of the migrations being applied.

Maybe we can take https://github.com/nautobot/nautobot-plugin-chatops/pull/76/files as an example (where it has logic in the migrations to do things differently for Postgres vs MySQL as `connection.vendor`?) Sorry to bring this up so late in this review.

chadell:
> Good to see that it applies the `11_mysql` migration but I'm concerned that the updates made to the earlier migrations aren't getting applied. I suspect that if you check the DB after this migration, for instance, `Note.title` is still a TextField (`varchar`?) rather than a 512-character-max-length CharField, which is different than you'll probably see if you create the DB fresh from scratch with all of the migrations being applied.
> 
> Maybe we can take https://github.com/nautobot/nautobot-plugin-chatops/pull/76/files as an example (where it has logic in the migrations to do things differently for Postgres vs MySQL as `connection.vendor`?) Sorry to bring this up so late in this review.

yes, I also think you are right, and I should handle the exceptions by `connection.vendor

chadell:
@glennmatthews I used the approach you suggested in ea2dab1e29a44501e88de93b760fd4e9c56ce9f1 
Does this looks better?

glennmatthews (approved):
Looks good to me. Thanks for fixing the migrations!

PR #163, opened 2021-11-29
Always refresh token when calling test_authentication(); add admin to delete NotificationSources

### LINKED ISSUE #162: "Validate authentication" button doesn't actually validate authentication for GmailAPIOauth
### Environment

- Python version: 
- Nautobot version: 1.1.5

### Expected Behavior

If access to a Gmail OAuth account is revoked, clicking the `Validate Authentication` button for that Notification Source should re-authenticate and re-request access to the account.

### Observed Behavior

`Validate Authentication` reports "Authentication Status SUCCESS: Test OK", even though actually attempting to retrieve records from this notification source will fail with ` ('invalid_grant: Token has been expired or revoked.', {'error': 'invalid_grant', 'error_description': 'Token has been expired or revoked.'})`

### Steps to Reproduce

1. Set up an OAuth account and successfully retrieve notifications from this account
2. In the Google Account settings, revoke access to the Nautobot application
3. In Nautobot, run "Validate Authentication" and see that it reports success still
4. In Nautobot, run a notification sync and see that it fails.

The issue appears to be with the following code chain:

- `NotificationSourceValidate` view calls `source.test_authentication()`
- `Source.test_authentication()` calls `self._authentication_logic()`
- `GmailAPI._authentication_logic()` calls `self.load_credentials()`
- `GmailAPIOauth.load_credentials()` reads the existing authentication token from the DB and if it's flagged as valid, does nothing further to validate the token (no actual call to the Google server)

glennmatthews:
Fixes #162, in theory (as usual it's really not feasible to actually test the OAuth workflow locally).

- Add capability in the Admin UI to manually delete NotificationSource records, forcing their recreation when Nautobot is restarted (useful in the case where a NotificationSource has a bad `token` and it's not feasible to access `nbshell`)
- Add a `force_refresh` flag to the `load_credentials()` method and set this flag to `True` in the case of `test_authentication` calling it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 # Changelog
 
-## v0.4.1
+## v0.4.2
+
+### Fixed
+
+- #162 - Fix inability to refresh OAuth authentication before token expires
+
+## v0.4.1 - 2021-11-29
 
 ### Fixed
 
--- a/nautobot_circuit_maintenance/admin.py
+++ b/nautobot_circuit_maintenance/admin.py
@@ -1,0 +1,17 @@
+"""Admin management content for this plugin."""
+
+from django.contrib import admin
+from .models import NotificationSource
+
+
+@admin.register(NotificationSource)
+class NotificationSourceAdmin(admin.ModelAdmin):
+    """Admin capabilities for viewing and deleting NotificationSource records."""
+
+    def has_add_permission(self, request):
+        """No option to add NotificationSources via the admin - use nautobot_config.py definitions instead."""
+        return False
+
+    def has_change_permission(self, request, obj=None):
+        """No option to change NotificationSources via the admin - use nautobot_config.py definitions instead."""
+        return False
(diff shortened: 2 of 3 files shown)


chadell (dismissed):
LGTM (could you update changelog?)

chadell:
> Fixes #162, in theory (as usual it's really not feasible to actually test the OAuth workflow locally).

do you think it would be worthy to implement proper mocking for testing OAuth?

glennmatthews:
> do you think it would be worthy to implement proper mocking for testing OAuth?

To be honest, I'm not sure what that would look like here - the API calls for OAuth are pretty abstract (`self.credentials.refresh(Request())` being the main one) so there's not much to mock unless we want to dig into the guts of the google.oauth2 library.

chadell:
> > do you think it would be worthy to implement proper mocking for testing OAuth?
> 
> To be honest, I'm not sure what that would look like here - the API calls for OAuth are pretty abstract (`self.credentials.refresh(Request())` being the main one) so there's not much to mock unless we want to dig into the guts of the google.oauth2 library.

I agree

PR #166, opened 2021-12-06
Handle NO-CHANGE status and unknown status values

glennmatthews:
Counterpart to https://github.com/networktocode/circuit-maintenance-parser/pull/125:

- if the parser library reports a status of `NO-CHANGE`, do not update the status of an existing CircuitMaintenance.
- if the parser library reports a status value that we don't understand (including `NO-CHANGE` for the first notification we see for a given maintenance), set the CircuitMaintenance status to `UNKNOWN`.
- Bump the version to 0.4.2 in preparation for an upcoming release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixed
 
 - #162 - Fix inability to refresh OAuth authentication before token expires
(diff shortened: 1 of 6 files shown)


glennmatthews:
Also I kept hitting an error during `invoke unittest` due to not loading the models properly during the initial signal handling, so I went ahead and fixed that as well.

chadell (approved):
LGTM

PR #168, opened 2021-12-06
Discard token if credentials refresh fails

glennmatthews:
Based on behavior I'm seeing in a live deployment, if the user revokes OAuth consent for the application, attempts to refresh the token raise a `RefreshError`. Added some logic to catch this exception and discard the saved token before reraising the exception.

As usual with OAuth, it's difficult to test this locally, but I'm reasonably confident this should do what it's intended to do.


chadell (dismissed):
LGTM, could you update Changelog?

chadell (approved):
LGTM

PR #173, opened 2021-12-10
Release v0.4.3

glennmatthews:
Only one fix here, but since a user is encountering this issue and it causes the entire job to blow up, it's kind of an important fix. :-)


chadell (approved):
LGTM

PR #176, opened 2021-12-16
View/template/table usability improvements

### LINKED ISSUE #165: Improvements to various views
### Environment

- Nautobot version: 1.1.5

### Proposed Functionality

`plugins/circuit-maintenance/maintenance/`:

1. Currently the table view defaults to listing maintenances in ascending chronological order, meaning the oldest maintenances are listed first and the newest last. I think it might be more user-friendly if  the newest (i.e. upcoming) maintenances were listed first instead.
2. Although there are options in the filter form to filter by Provider and by Circuit, there are no columns in the table to display these attributes.
     a. There's no visible logic to the ordering of Circuits in the drop-down.
     b. This should probably be changed from a `ModelChoiceField` to a `DynamicModelMultipleChoiceField`.
3. There's no filter option for filtering by Status, which would be very useful.

`plugins/circuit-maintenance/rawnotification/`:

1. As with the other table, this defaults to ascending order by stamp, showing oldest notifications first, but I think the reverse would be more logical as a default.
2. There's no filter option for filtering by Provider.

`plugins/circuit-maintenance/maintenance/<uuid>/`:

1. Would it be possible to add a direct link from this view to the Provider that this maintenance is associated with? (CircuitMaintenance -> ParsedNotification -> RawNotification -> Provider)

`plugins/circuit-maintenance/rawnotification/<uuid>/`:

1. The provider field in this view could be a hyperlink to the Provider model.
2. The source field in this view could be a hyperlink to the NotificationSource model.

### Use Case

General user experience

glennmatthews:
Fixes #165.

- Add a read-only REST API endpoint for NotificationSource objects (needed so that we can use this endpoint to populate a DynamicModelMultipleChoiceField in the RawNotificationFilterSetForm).
- Add filterset classes as available to the REST API viewsets and update them to subclass from CustomFieldModelViewSet since these models support custom fields. 
- Change CircuitMaintenanceFilterSet to use ModelMultipleChoiceFilter instead of CharFilter for filtering by provider or circuit
- Add provider and notification-source filters to RawNotificationFilterSet
- Add status filter to CircuitMaintenanceFilterForm, and change provider and circuit filters to DynamicModelMultipleChoiceFields
- Add provider and notification-source filters to RawNotificationFilterSetForm
- Add circuits and providers columns to CircuitMaintenanceTable
- Add linkification to many columns in various tables and fields in various detail views
- Add or correct breadcrumbs in various detail views
- Add "Notifying provider" to CircuitMaintenance detail view
- Change ordering of CircuitMaintenanceListView to show most recent maintenances first
- Change ordering of RawNotificationListView to show most recent notifications first

image

image

image

image

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## v0.5.0
+
+### Changed
+
+- #165 - Numerous UI/UX refinements and improvements. See #176 for details.
+
 ## v0.4.3
 
 ### Fixed
--- a/nautobot_circuit_maintenance/api/serializers.py
+++ b/nautobot_circuit_maintenance/api/serializers.py
@@ -1,7 +1,8 @@
 """Serializer for Circuit Maintenance API."""
 from rest_framework import serializers
+from nautobot.circuits.api.nested_serializers import NestedProviderSerializer
 from nautobot.circuits.api.serializers import CircuitSerializer
-from nautobot_circuit_maintenance.models import CircuitImpact, CircuitMaintenance, Note
+from nautobot_circuit_maintenance.models import CircuitImpact, CircuitMaintenance, Note, NotificationSource
 
 
 class CircuitMaintenanceCircuitImpactSerializer(serializers.ModelSerializer):
@@ -38,3 +39,18 @@ class NoteSerializer(serializers.ModelSerializer):
 
         model = Note
         fields = ["id", "maintenance", "title", "comment"]
+
+
+class NotificationSourceSerializer(serializers.ModelSerializer):
+    """Serializer for NotificationSource records."""
+
+    url = serializers.HyperlinkedIdentityField(
+        view_name="plugins-api:nautobot_circuit_maintenance-api:notificationsource-detail"
+    )
+    providers = NestedProviderSerializer(many=True)
+
+    class Meta:
+        """Meta class for NotificationSourceSerializer."""
+
+        model = NotificationSource
+        fields = ["id", "url", "name", "slug", "providers", "attach_all_providers"]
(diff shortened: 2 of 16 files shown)


chadell:
Wow, amazing work!

There is something I don't fully get. 
In the last example picture, for a CircuitMaintenance, how do we get multiple `Providers`? 🤔 
and what is the difference between "Notifying Providers" (empty) and the `Providers` from the `Circuits`?

chadell:
> > In the last example picture, for a CircuitMaintenance, how do we get multiple `Providers`? 🤔
> 
> In this case I manually created the CircuitMaintenance and its associated CircuitImpacts. There's nothing in the code to require that all CircuitImpacts for a CircuitMaintenance belong to the same Provider. (And I can think of plausible scenarios where a single maintenance window might impact multiple Providers' circuits.)

The way we [create](https://github.com/nautobot/nautobot-plugin-circuit-maintenance/blob/develop/nautobot_circuit_maintenance/handle_notifications/handler.py#L196) the composite CircuitMaintenance is using the `Provider.slug` from the Notification, so I guess the scenario you are thinking on is that from a CircuitMaintenace from a Provider, we could infer dependent Circuits that are transported by the original one? 

> 
> > and what is the difference between "Notifying Providers" (empty) and the `Providers` from the `Circuits`?
> 
> "Notifying Provider" is derived from the RawNotification(s) that link to this CircuitMaintenance. It's mostly useful for the case where a RawNotification references CIDs that aren't in Nautobot, so there are Notes created but no CircuitImpacts - I wanted an easy way to identify which provider the CircuitMaintenance came from in that case.

Got it. In this example is empty because you created the CircuitImpact manually 😄

chadell (dismissed):
LGTM

chadell:
could you update Changelog?

chadell (approved):
LGTM

PR #195, opened 2022-07-09
Updates to Drop Python 3.6

jvanderaa:
Looking to drop the Python3.6 compatibility. 

- Drops end of support Python version
- Supports M1 Mac Support
- Nautobot no longer supports Python3.6 with the new features

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,7 +1,4 @@
 [flake8]
 # E501: Line length is enforced by Black, so flake8 doesn't need to check it
 # W503: Black disagrees with this rule, as does PEP 8; Black wins
-# E402: Module level import not at top of file, due to importlib check
-# E701: Flake8 Catches errors in data model
-ignore = E501, W503, E701
-exclude = nautobot_circuit_maintenance/__init__.py
+ignore = E501, W503
--- a/.flake8
+++ b/.flake8
@@ -1,4 +1,7 @@
 [flake8]
 # E501: Line length is enforced by Black, so flake8 doesn't need to check it
 # W503: Black disagrees with this rule, as does PEP 8; Black wins
-ignore = E501, W503
+# E402: Module level import not at top of file, due to importlib check
+# E701: Flake8 Catches errors in data model
+ignore = E501, W503, E701
+exclude = nautobot_circuit_maintenance/__init__.py
--- a/.flake8
+++ b/.flake8
@@ -1,4 +1,5 @@
 [flake8]
 # E501: Line length is enforced by Black, so flake8 doesn't need to check it
 # W503: Black disagrees with this rule, as does PEP 8; Black wins
-ignore = E501, W503
+# E701: Colon multi-line indicator, but is just used for model definition. Unable to fix in file.
+ignore = E501, W503, E701
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -103,7 +103,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9"]
+        python-version: ["3.7", "3.8", "3.9", "3.10"]
         nautobot-version: ["1.1.5"]
     runs-on: "ubuntu-20.04"
     env:
@@ -126,7 +126,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9"]
+        python-version: ["3.7", "3.8", "3.9", "3.10"]
         nautobot-version: ["1.1.5"]
     runs-on: "ubuntu-20.04"
     env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,7 +104,12 @@ jobs:
       fail-fast: true
       matrix:
         python-version: ["3.7", "3.8", "3.9"]
-        nautobot-version: ["1.1.5", "1.2.11", "1.3.7"]
+        nautobot-version: ["stable"]
+        db-backend: ["postegresql"]
+        include:
+          - python-version: "3.9"
+            db-backend: "postgresql"
+            nautobot-version: "1.3.3"
     runs-on: "ubuntu-20.04"
     env:
       INVOKE_NAUTOBOT_CIRCUIT_MAINTENANCE_PYTHON_VER: "${{ matrix.python-version }}"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -132,7 +132,10 @@ jobs:
       fail-fast: true
       matrix:
         python-version: ["3.7", "3.8", "3.9"]
-        nautobot-version: ["1.1.5", "1.2.11", "1.3.7"]
+        nautobot-version: ["1.3.8"]
+        include:
+          - python-version: "3.9"
+            nautobot-version: "1.2.11"
     runs-on: "ubuntu-20.04"
     env:
       INVOKE_NAUTOBOT_CIRCUIT_MAINTENANCE_PYTHON_VER: "${{ matrix.python-version }}"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,12 +104,12 @@ jobs:
       fail-fast: true
       matrix:
         python-version: ["3.7", "3.8", "3.9"]
-        nautobot-version: ["stable"]
+        nautobot-version: ["1.3.8"]
         db-backend: ["postegresql"]
         include:
           - python-version: "3.9"
             db-backend: "postgresql"
-            nautobot-version: "1.3.3"
+            nautobot-version: "1.2.11"
     runs-on: "ubuntu-20.04"
     env:
       INVOKE_NAUTOBOT_CIRCUIT_MAINTENANCE_PYTHON_VER: "${{ matrix.python-version }}"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -103,7 +103,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        python-version: ["3.7", "3.8", "3.9", "3.10"]
+        python-version: ["3.7", "3.8", "3.9"]
         nautobot-version: ["1.1.5"]
     runs-on: "ubuntu-20.04"
     env:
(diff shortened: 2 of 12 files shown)


jvanderaa:
Hmm, trying to figure out why the Flake8 tests are failing. They are supposed to be ignored.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "nautobot-circuit-maintenance"
-version = "0.5.0b1"
+version = "0.5.0"
```
glennmatthews, line 3: Let's not automatically bump the release version in every PR to develop. I see there are 5 other open PRs at present - updating the release version should be done when we're ready to cut a new release and only then.

pyproject.toml
```diff
@@ -44,7 +44,9 @@ bandit = "*"
 pylint = "*"
```
glennmatthews, line 44: Can we remove the pinning to a specific `typed-ast` version above? Or is this still an issue with black?

pyproject.toml
```diff
@@ -44,7 +44,9 @@ bandit = "*"
...
 pylint-django = "*"
 pydocstyle = "*"
-flake8 = "*"
+# we need to pin flake8 because of package dependencies that cause it to downgrade and 
+# therefore cause issues with linting since older versions do not take .flake8 as config
+flake8 = "^3.9.2"
```
glennmatthews, line 49: This is a different approach than we've taken in other repos in this organization, where the fix there was to explicitly invoke flake8 as `flake8 --config .flake8`, e.g. https://github.com/nautobot/nautobot-plugin-netbox-importer/pull/75/files - is that alternative preferable?

Also, what dependencies do we have that were causing a downgrade to older versions of flake8?

.github/workflows/ci.yml
```diff
@@ -103,7 +103,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9"]
+        python-version: ["3.7", "3.8", "3.9"]
         nautobot-version: ["1.1.5"]
```
glennmatthews, line 107: Note that CI is *not* actually testing nautobot 1.1.5 at present, rather it's testing the version we have in `poetry.lock` which is now 1.3.7 - see https://github.com/nautobot/nautobot-plugin-circuit-maintenance/runs/7265693745?check_suite_focus=true#step:5:339 for example.

If we want CI to actually test the specified Nautobot version (and we do!) then we need to ensure that our Dockerfile does a `poetry add nautobot==$NAUTOBOT_VER` before it does a `poetry install`.

glennmatthews (changes requested):

pyproject.toml
jvanderaa: Fixing.

nautobot_circuit_maintenance/tests/test_views.py
```diff
@@ -69,6 +70,10 @@ def _get_base_url(self):
     def assertInstanceEqual(self, instance, data, api=False):  # pylint: disable=arguments-differ
         """Used to overwrite inbuilt function. Causing type issues for datetimepicker."""
 
+    @skip
+    def test_has_advanced_tag(self):
```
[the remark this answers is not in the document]
jvanderaa: This may have to come sooner than later. The skip tag isn't really skipping the test. It is still failing.

pyproject.toml
```diff
@@ -98,6 +96,7 @@ disable = """,
     too-few-public-methods,
     duplicate-code,
     too-many-ancestors,
+    arguments-differ,
```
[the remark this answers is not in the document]
jvanderaa: Done.

.flake8
```diff
@@ -1,4 +1,7 @@
 [flake8]
 # E501: Line length is enforced by Black, so flake8 doesn't need to check it
 # W503: Black disagrees with this rule, as does PEP 8; Black wins
-ignore = E501, W503
+# E402: Module level import not at top of file, due to importlib check
+# E701: Flake8 Catches errors in data model
+ignore = E501, W503, E701
+exclude = nautobot_circuit_maintenance/__init__.py
```
glennmatthews, line 7: That's a fair amount of code to just disable flake8 for across the board - same here, can we add targeted `# noqa` to specific lines instead of excluding the whole file?

scetron (commented):
Thanks @jvanderaa. Are the 'test_has_advanced_tab' now required or are these just place holders?

jvanderaa:
The tests are auto generated as part of the inheritance for the test case. We need to adjust the HTML templates to have a new base to get these views for free. @scetron

.flake8
jvanderaa: This list was quite larger number. Trying to get this squared away.

.github/workflows/ci.yml
```diff
@@ -126,8 +126,8 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9"]
-        nautobot-version: ["1.1.5"]
+        python-version: ["3.7", "3.8", "3.9"]
+        nautobot-version: ["1.1.5", "1.2.11", "1.3.7"]
```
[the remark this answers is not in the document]
jvanderaa: Working to get `stable` working here. The tests are not working, I need to dig into the parsing.

.github/workflows/ci.yml
```diff
@@ -132,7 +132,10 @@ jobs:
       fail-fast: true
       matrix:
         python-version: ["3.7", "3.8", "3.9"]
-        nautobot-version: ["1.1.5", "1.2.11", "1.3.7"]
+        nautobot-version: ["1.3.8"]
```
chadell, line 135: which was the problem with mysql?
BTW, would make sense to leave the nautboto-version to "stable" to always check against hte latest by default? (don't recall if we already commented this point)

jvanderaa: `stable` doesn't work. There is a parser error.

jvanderaa: I'm just trying to get things into a working state. What didn't work with MySQL is that I didn't have a matrix set up for it.

chadell: sorry Josh, I don't get it, because the matrix was already set up before, no?

development/Dockerfile
```diff
@@ -1,15 +1,16 @@
-ARG NAUTOBOT_VER="1.0.1"
+ARG NAUTOBOT_VER="1.3.5"
 ARG PYTHON_VER=3.8
```
scetron, line 2: I am wondering if there is a way to tokenize these so we can have them in one place?

.github/workflows/ci.yml
jvanderaa: Not for `stable`. It was set up to test individual versions.

.github/workflows/ci.yml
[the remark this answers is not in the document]
jvanderaa: @chadell this is what things looked like before.

.github/workflows/ci.yml
chadell: yes, I understand, not for stable, but the matrix was working well for mysql before, no?

jvanderaa: Yes it was. I guess the wording of `fix mysql` was poor. Probably should have just been `update mysql matrix`.

development/Dockerfile
jvanderaa: If it is, I would like that to be a separate PR. This is just to drop 3.6, not to improve CI.

chadell (approved):
to not block you, let's move this forward, but I would support adding the workaround for the `stable` that @glennmatthews suggested (in another PR if you want)

development/Dockerfile
scetron: Agreed.

jvanderaa:
I am not able to merge. I think this is still you @scetron and @chadell

nautobot_circuit_maintenance/tests/test_graphql.py
```diff
@@ -18,7 +18,7 @@ class GraphQLTestCase(TestCase):
     """GraphQL tests."""
 
     @classmethod
-    def setUp(cls):
+    def setUp(cls):  # pylint: disable=arguments-differ
```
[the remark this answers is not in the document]
chadell: @jvanderaa have you taken care of this?

scetron (dismissed):

nautobot_circuit_maintenance/tests/test_graphql.py
glennmatthews, line 21: remove `@classmethod`, change signature to `def setUp(self)`, pylint disable should no longer be needed.

nautobot_circuit_maintenance/tests/test_views.py
```diff
@@ -29,6 +30,11 @@ def _get_base_url(self):
     def assertInstanceEqual(self, instance, data, api=False):  # pylint: disable=arguments-differ
         """Used to overwrite inbuilt function. Causing type issues for datetimepicker."""
 
+    @skip
+    def test_has_advanced_tab(self):
+        """To remove tests"""
+        self.assertTrue()  # pylint: disable=no-value-for-parameter
```
glennmatthews, line 36: Can we make all of these consistent as simply:

```python
    @skip("Not implemented yet.")
    def test_has_advanced_tab(self):
        pass
```

glennmatthews (approved):

nautobot_circuit_maintenance/tests/test_graphql.py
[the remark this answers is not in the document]
jvanderaa: What is the thought? I am not clear in understanding what either is doing immediately. Just disabling pylint rule from the pylint updates.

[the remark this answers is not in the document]
jvanderaa: If not I will take a look more upon my return.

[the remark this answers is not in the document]
glennmatthews: `@classmethod setUpTestData(cls)` runs once before all tests in this TestCase, `setUp(self)` runs before each individual test in this TestCase. `@classmethod setUp(cls)` is some sort of weird hybrid between these two and is incorrect. I suspect that changing it to `setUp(self)` should be the correct fix here.

glennmatthews (approved):

Issue #177, opened 2021-12-21
Add Edit/Delete buttons to Note Detail UI view


chadell:
### Environment

- Nautobot version: 
- nautobot-circuit-maintenance version: 0.4.3

### Proposed Functionality

Add Edit and Delete buttons in the Note Detail view

### Use Case

Easier access to these actions

chadell:
won't do because notes will be deprecated to use Nautobot default ones

PR #201, opened 2022-08-17
Move to nautobot base templates

### LINKED ISSUE #196: Switch to Object Detail View
### Environment

- Nautobot version: 1.3.8
- nautobot-circuit-maintenance version: 0.5.0

### Proposed Functionality

Migrate detail views to object detail view inherited template

### Use Case

Get all of the goodness and consistency from using a base template with more details.

scetron:
This adopts the newer Nautobot base templates and fixes other minor issues with templates.

Fixes #196 and #177


itdependsnetworks:
I think you need to bump this as well https://github.com/nautobot/nautobot-plugin-circuit-maintenance/blob/develop/nautobot_circuit_maintenance/__init__.py#L86

chadell:
are these new templates available in all the Nautobot versions we support in the plugin? I mean, I know we make Nautobot Core backwards compatible, but maybe the new templates were not available before, so the plugin won't work in older versions of Nautobot.
I know there was a conversation about this few days ago. For instance, in Golden Config, there is a table of the minimum supported Nautobot version for the plugin version. Maybe we should do something like this here, instead of trying to keep full backwards compatibility in the plugin (keeping both templates versions, for instance)

itdependsnetworks:
What we came to the conclusion on GC was. Let's not restrict (below a Nautobot 2.0) but let's be aggressive in bumping versions to not support older versions of Nautobot. This allows us to not have to worry about backwards compatibility and by logical sense, anything that breaks with the legacy versions is a Nautobot Core concern. Documented the approach here: https://github.com/nautobot/nautobot-plugin-golden-config/blob/develop/docs/admin/admin_install.md#deprecation-policy

chadell:
Yes, I also agree that creating a well documented deprecation policy could grant us an easier development moving forward.
So, why not start on this PR? Plugin `>=0.6.0 only supports Nautobot 1.4.0`

jvanderaa:
@chadell should we get this implemented before the 1.0.0? And possibly look to add `nautobot.utilities.filters.MultiValueDateTimeFilter` to the search?

chadell:
Isn't most of it covered by https://github.com/nautobot/nautobot-plugin-circuit-maintenance/commit/0f4a3a23f3e96107281a4d6fd1e2ad541f9c2217 ?

For support `nautobot.utilities.filters.MultiValueDateTimeFilter`, we can tackle it as an improvement for a future minor release.

jvanderaa:
Looks like we have a lot out of date now @scetron

chadell:
closing as with Nautobot 2.4, the new AppView pattern will be applied

Issue #213, opened 2022-09-24
Job Creation: Search for Sites Without Redundant Circuit


jvanderaa:
### Environment

- Nautobot version: 
- nautobot-circuit-maintenance version: 

### Proposed Functionality

I'm looking to add a Nautobot Job to get packaged that will search through the possible records, and flag any sites that will not have a redundant circuit available during a maintenance of another circuit.

```
        # Query for all of the circuits maintenances that are on going in the future
        # Query for all of the circuits within Nautobot
        # Loop over each of the circuit maintenance records
            # Check to see if there are any circuit maintenances that are duplicated time
            # Report failures for any time where a circuit will take an outage
            # Evaluate adding a tag of an impeding site outage due to WAN outages
            # Log success for each time there is a known circuit still available at the site at the same time
```

### Use Case

Give the ability to notify when there is an outage being created by maintenance events, even though the network design has redundancy of circuits.

chadell:
is this addressed by #214?

PR #214, opened 2022-09-26
Maintenance Overlap Job Addition

jvanderaa:
This adds a job to search for overlapping maintenances for circuits at a site.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ PLUGINS_CONFIG = {
         "raw_notification_initial_days_since": 100,
         "raw_notification_size": 16384,
         "dashboard_n_days": 30,  # Defaults to 30 days in the configurations, change/override here
-        "overlap_job_log_only_impacts": False, # Defaults for False to alert on any notification, regardless of impact. True if there needs to be a known impact.
+        "overlap_job_exclude_no_impact": False, # Exclude in job warnings the impact of `No-Impact`
         "notification_sources": [
             {
               ...
--- a/README.md
+++ b/README.md
@@ -33,6 +33,7 @@ PLUGINS_CONFIG = {
         "raw_notification_initial_days_since": 100,
         "raw_notification_size": 16384,
         "dashboard_n_days": 30,  # Defaults to 30 days in the configurations, change/override here
+        "overlap_job_log_only_impacts": False, # Defaults for False to alert on any notification, regardless of impact. True if there needs to be a known impact.
         "notification_sources": [
             {
               ...
@@ -215,41 +216,6 @@ In the Circuit Maintenance plugin UI section, there is a **Notification Sources*
 Because the Notification Sources are defined by the configuration, you can only view and edit `providers`, but not `add` or `delete` new Notification Sources via UI or API.
 
 > Note that for emails from a given Provider to be processed, you must _both_ define a source email address(es) for that Provider (Usage section 1, above) _and_ add that provider to a specific Notification Source as described in this section.
-
-### 3. Run Handle Notifications Job
-
-There is an asynchronous task defined as a **Nautobot Job**, **Handle Circuit Maintenance Notifications** that will connect to the emails sources defined under the Notification Sources section (step above), and will fetch new notifications received since the last notification was fetched.
-Each notification will be parsed using the [circuit-maintenance-parser](https://github.com/networktocode/circuit-maintenance-parser) library, and if a valid parsing is executed, a new **Circuit Maintenance** will be created, or if it was already created, it will updated with the new data.
-
-So, for each email notification received, several objects will be created:
-
-#### 3.1 Notification
-
-Each notification received will create a related object, containing the raw data received, and linking to the corresponding **parsed notification** in case the [circuit-maintenance-parser](https://github.com/networktocode/circuit-maintenance-parser) was able to parse it correctly.
-
-#### 3.2 Parsed Notification
-
-When a notification was successfully parsed, it will create a **parsed notification** object, that will contain the structured output from the parser library , following the recommendation defined in [draft NANOG BCOP](https://github.com/jda/maintnote-std/blob/master/standard.md), and a link to the related **Circuit Maintenance** object created.
-
-#### 3.3 Circuit Maintenance
-
-The **Circuit Maintenance** is where all the necessary information related to a Circuit maintenance is tracked, and reuses most of the data model defined in the parser library.
-
-Attributes:
-
-- Name: name or identifier of the maintenance.
-- Description: description of the maintenance.
-- Status: current state of the maintenance.
-- Start time: defined start time of the maintenance work.
-- End time: defined end time of the maintenance work.
-- Ack: boolean to show if the maintenance has been properly handled by the operator.
-- Circuits: list of circuits and its specific impact linked to this maintenance.
-- Notes: list of internal notes linked to this maintenance.
-- Notifications: list of all the parsed notifications that have been processed for this maintenance.
-
-<p align="center">
-<img src="https://raw.githubusercontent.com/nautobot/nautobot-plugin-circuit-maintenance/develop/docs/images/circuit_maintenance.png" width="800" class="center">
-</p>
 
 ### Metrics
 
--- a/docs/jobs.md
+++ b/docs/jobs.md
@@ -35,6 +35,6 @@ Attributes:
 
 ## Circuit Overlap Job
 
-The circuit overlap job that gets included with the Circuit Maintenance Plugin is a job that is going to search for possible overlapping maintenances, which **may** cause an outage of a site. The variable `overlap_job_log_only_impacts` controls on the check if a maintenance notification has an expected impact. Default is `False` for this setting, that any maintenance notification will be alerted on within the Nautobot Job. 
+The circuit overlap job that gets included with the Circuit Maintenance Plugin is a job that is going to search for possible overlapping maintenances, which **may** cause an outage of a site. The variable `overlap_job_exclude_no_impact ` controls on the check if a maintenance notification has an expected impact. Default is `False` for this setting, that any maintenance notification will be alerted on within the Nautobot Job. 
 
 Use the Job regularly to search for the overlap and review any log message that has a Warning level that will indicate that there is a possible overlapping maintenance.
--- a/docs/jobs.md
+++ b/docs/jobs.md
@@ -1,0 +1,40 @@
+# Jobs
+
+## Run Handle Notifications Job
+
+There is an asynchronous task defined as a **Nautobot Job**, **Handle Circuit Maintenance Notifications** that will connect to the emails sources defined under the Notification Sources section (step above), and will fetch new notifications received since the last notification was fetched.
+Each notification will be parsed using the [circuit-maintenance-parser](https://github.com/networktocode/circuit-maintenance-parser) library, and if a valid parsing is executed, a new **Circuit Maintenance** will be created, or if it was already created, it will updated with the new data.
+
+So, for each email notification received, several objects will be created:
+
+### Notification
+
+Each notification received will create a related object, containing the raw data received, and linking to the corresponding **parsed notification** in case the [circuit-maintenance-parser](https://github.com/networktocode/circuit-maintenance-parser) was able to parse it correctly.
+
+### Parsed Notification
+
+When a notification was successfully parsed, it will create a **parsed notification** object, that will contain the structured output from the parser library , following the recommendation defined in [draft NANOG BCOP](https://github.com/jda/maintnote-std/blob/master/standard.md), and a link to the related **Circuit Maintenance** object created.
+
+### Circuit Maintenance
+
+The **Circuit Maintenance** is where all the necessary information related to a Circuit maintenance is tracked, and reuses most of the data model defined in the parser library.
+
+Attributes:
+
+- Name: name or identifier of the maintenance.
+- Description: description of the maintenance.
+- Status: current state of the maintenance.
+- Start time: defined start time of the maintenance work.
+- End time: defined end time of the maintenance work.
+- Ack: boolean to show if the maintenance has been properly handled by the operator.
+- Circuits: list of circuits and its specific impact linked to this maintenance.
+- Notes: list of internal notes linked to this maintenance.
+- Notifications: list of all the parsed notifications that have been processed for this maintenance.
+
+![Circuit Maintenance Job](images/circuit_maintenance.png)
+
+## Circuit Overlap Job
+
+The circuit overlap job that gets included with the Circuit Maintenance Plugin is a job that is going to search for possible overlapping maintenances, which **may** cause an outage of a site. The variable `overlap_job_log_only_impacts` controls on the check if a maintenance notification has an expected impact. Default is `False` for this setting, that any maintenance notification will be alerted on within the Nautobot Job. 
+
+Use the Job regularly to search for the overlap and review any log message that has a Warning level that will indicate that there is a possible overlapping maintenance.
--- a/nautobot_circuit_maintenance/__init__.py
+++ b/nautobot_circuit_maintenance/__init__.py
@@ -90,7 +90,7 @@ class CircuitMaintenanceConfig(PluginConfig):
         "raw_notification_initial_days_since": 7,
         "raw_notification_size": 8192,
         "dashboard_n_days": 30,
-        "overlap_job_log_only_impacts": False,
+        "overlap_job_exclude_no_impact": False,
     }
     caching_config = {}
     home_view_name = "plugins:nautobot_circuit_maintenance:circuitmaintenance_overview"
--- a/nautobot_circuit_maintenance/__init__.py
+++ b/nautobot_circuit_maintenance/__init__.py
@@ -90,6 +90,7 @@ class CircuitMaintenanceConfig(PluginConfig):
         "raw_notification_initial_days_since": 7,
         "raw_notification_size": 8192,
         "dashboard_n_days": 30,
+        "overlap_job_log_only_impacts": False,
     }
     caching_config = {}
     home_view_name = "plugins:nautobot_circuit_maintenance:circuitmaintenance_overview"
--- a/nautobot_circuit_maintenance/jobs/__init__.py
+++ b/nautobot_circuit_maintenance/jobs/__init__.py
@@ -1,0 +1,8 @@
+"""Circuit Maintenance plugin jobs."""
+# from nautobot.extras.models import Job
+
+from nautobot_circuit_maintenance.handle_notifications.handler import HandleCircuitMaintenanceNotifications
+from nautobot_circuit_maintenance.jobs.site_search import FindSitesWithCircuitImpact
+
+
+jobs = [FindSitesWithCircuitImpact, HandleCircuitMaintenanceNotifications]
--- a/nautobot_circuit_maintenance/jobs/__init__.py
+++ b/nautobot_circuit_maintenance/jobs/__init__.py
@@ -1,6 +1,5 @@
 """Circuit Maintenance plugin jobs."""
-# from nautobot.extras.models import Job
-
+# pylint: disable=no-name-in-module
 from nautobot_circuit_maintenance.handle_notifications.handler import HandleCircuitMaintenanceNotifications
 from nautobot_circuit_maintenance.jobs.site_search import FindSitesWithCircuitImpact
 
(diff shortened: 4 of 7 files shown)


jvanderaa:
Fixes #213

nautobot_circuit_maintenance/jobs/__init__.py
```diff
@@ -0,0 +1,7 @@
+"""Circuit Maintenance plugin jobs."""
+# pylint: disable=no-name-in-module
+from nautobot_circuit_maintenance.handle_notifications.handler import HandleCircuitMaintenanceNotifications
+from nautobot_circuit_maintenance.jobs.site_search import FindSitesWithCircuitImpact
+
+
+jobs = [FindSitesWithCircuitImpact, HandleCircuitMaintenanceNotifications]
```
chadell, line 7: and the original `nautobot_circuit_maintenance/jobs.py`? it should be removed

nautobot_circuit_maintenance/jobs/site_search.py
```diff
@@ -0,0 +1,161 @@
...
+from typing import NamedTuple
+
+from django.conf import settings
+
+from nautobot.extras.jobs import Job
+from nautobot.circuits.models import Circuit
+
+from nautobot_circuit_maintenance.models import CircuitMaintenance
+
+name = "Circuit Maintenance"  # pylint: disable=invalid-name
```
chadell, line 13: don't understand what is this used for

nautobot_circuit_maintenance/jobs/site_search.py
```diff
@@ -0,0 +1,161 @@
...
+PLUGIN_SETTINGS = settings.PLUGINS_CONFIG.get("nautobot_circuit_maintenance", {})
+CIRCUIT_MAINTENANCE_TAG_COLOR = "Purple"
+
+
+class Range(NamedTuple):
+    """Range class for use within the check for overlap.
+
+    Args:
```
chadell, line 21: `NamedTuple` is the parent class

nautobot_circuit_maintenance/jobs/site_search.py
```diff
@@ -0,0 +1,161 @@
...
+from nautobot.circuits.models import Circuit
+
+from nautobot_circuit_maintenance.models import CircuitMaintenance
+
+name = "Circuit Maintenance"  # pylint: disable=invalid-name
+PLUGIN_SETTINGS = settings.PLUGINS_CONFIG.get("nautobot_circuit_maintenance", {})
+CIRCUIT_MAINTENANCE_TAG_COLOR = "Purple"
+
+
+class Range(NamedTuple):
```
chadell, line 18: maybe a more specific class name? `DateTimeRange` ?

nautobot_circuit_maintenance/jobs/site_search.py
```diff
@@ -0,0 +1,161 @@
...
+    latest_start = max(range1.start, range2.start)
+    earliest_end = min(range1.end, range2.end)
+
+    # Get the delta, compare versus 0 (negative number is no overlap)
+    delta = (earliest_end - latest_start).days + 1
+
+    # If delta is negative there is no overlap, return false
+    overlap = max(0, delta)
+
+    return overlap != 0
```
chadell, line 53: I believe a more explicit logic should be possible...

nautobot_circuit_maintenance/jobs/site_search.py
```diff
@@ -0,0 +1,161 @@
...
+    Args:
+        circuit (Circuit): Nautobot Circuit Object to find the sites
+
+
+    Returns:
+        list: List of Nautobot Site objects
+    """
+    site_list: list = []
+    for term in [circuit.termination_a, circuit.termination_z]:
+        if term is not None and getattr(term, "provider_network") is None and term.site not in site_list:
```
chadell, line 68: a `set` would simplify this logic

nautobot_circuit_maintenance/jobs/site_search.py
```diff
@@ -0,0 +1,161 @@
...
+        for circuit in record.circuits:
+            # Check both termination A and Z for values of None, add that record to the set for each site
+            for term in [circuit.termination_a, circuit.termination_z]:
+                if term is not None:
+                    return_dictionary[term.site.name].add(record)
+
+    return dict(return_dictionary)
+
+
+class FindSitesWithCircuitImpact(Job):
```
chadell, line 111: CircuitImpact has already a specific meaning in this plugin, maybe another name would make it clearer

nautobot_circuit_maintenance/jobs/site_search.py
```diff
@@ -0,0 +1,161 @@
...
+    """Nautobot Job definition for finding sites without redundant circuit for impactful maintenance.
+
+    Args:
+        Job (Nautobot Job): Nautobot Job import.
```
chadell, line 115: Job is the parent class

nautobot_circuit_maintenance/jobs/site_search.py
```diff
@@ -0,0 +1,161 @@
...
+    """
+
+    class Meta:
+        """Meta definition for the Job."""
+
+        name = "Find Sites With Circuit Impact"
```
chadell, line 121: check note about `CircuitImpact`

nautobot_circuit_maintenance/jobs/site_search.py
```diff
@@ -0,0 +1,161 @@
...
+        # Build a circuit mapper
+        circuit_maintenance_mapper = build_sites_to_maintenance_mapper(circuit_maintenances)
+
+        # Loop over each of the circuit maintenance records
+        for ckt_maint in circuit_maintenances:
+            # Get the list of sites
+            for circuit in ckt_maint.circuits:
+                sites = get_sites_from_circuit(circuit)
+
+            # Check to see if there are any circuit maintenances that are duplicated time
```
chadell, line 145:
```suggestion
            # Check to see if there are any circuit maintenances that are overlapping at some moment in time
```

nautobot_circuit_maintenance/jobs/site_search.py
```diff
@@ -0,0 +1,161 @@
...
+        """
+        # Query for all of the circuits maintenances that are on going in the future
+        today = date.today()
+        circuit_maintenances = CircuitMaintenance.objects.filter(start_time__gte=today).order_by("start_time")
+
+        # Build a circuit mapper
+        circuit_maintenance_mapper = build_sites_to_maintenance_mapper(circuit_maintenances)
+
+        # Loop over each of the circuit maintenance records
+        for ckt_maint in circuit_maintenances:
```
chadell, line 140: `ckt_maint`? I don't get what `ckt` means

nautobot_circuit_maintenance/jobs/site_search.py
```diff
@@ -0,0 +1,161 @@
...
+            # Get the list of sites
+            for circuit in ckt_maint.circuits:
+                sites = get_sites_from_circuit(circuit)
```
chadell, line 143:
```suggestion
                sites += get_sites_from_circuit(circuit)
```
and define `sites = []` outside of the loop

nautobot_circuit_maintenance/jobs/site_search.py
```diff
@@ -0,0 +1,161 @@
...
+            for site in sites:
+                for other_ckt_maint in circuit_maintenance_mapper[site.name]:
+                    # Report failures for any time where a circuit will take an outage
+                    if ckt_maint == other_ckt_maint:
+                        self.log_info("Found each other!")
+                        continue
+                    if check_for_overlap(ckt_maint, other_ckt_maint):
+                        self.log_warning(
+                            obj=site,
+                            message=f"There is an overlapping maintenance for site: {site.name}. Other maintenances: {other_ckt_maint}",
```
chadell, line 155: should not document both maintenances? why only one?

nautobot_circuit_maintenance/jobs/site_search.py
```diff
@@ -0,0 +1,161 @@
...
+                        )
+                    else:
+                        # Log success for each time there is a known circuit still available at the site at the same time
+                        self.log_success(
```
chadell, line 159: I would leave this as `debug`, I believe this should be the "default"

nautobot_circuit_maintenance/tests/test_jobs.py
```diff
@@ -0,0 +1,69 @@
+"""Tests for Jobs being included."""
```
chadell, line 1: I would make this test specific to the site_search job

nautobot_circuit_maintenance/tests/test_jobs.py
```diff
@@ -0,0 +1,69 @@
...
+from typing import NamedTuple
+
+from django.test import TestCase
+
+from nautobot_circuit_maintenance.jobs.site_search import check_for_overlap
+
+DATE_FORMAT = "%Y-%m-%d %H:%M:%S"
+
+
+class MockRecord(NamedTuple):
```
chadell, line 13: why not import the `class Range(NamedTuple)` ?

nautobot_circuit_maintenance/jobs/site_search.py
```diff
@@ -0,0 +1,161 @@
...
+    """
+    return_dictionary = collections.defaultdict(set)
+    for record in maintenance_queryset:
+        # Get the site overall
+        # Start with looping over the circuits that are defined
+        for circuit in record.circuits:
+            # Check both termination A and Z for values of None, add that record to the set for each site
+            for term in [circuit.termination_a, circuit.termination_z]:
+                if term is not None:
+                    return_dictionary[term.site.name].add(record)
```
chadell, line 106: we should not include circuit which circuit_impact is `NO_IMPACT`
https://github.com/networktocode/circuit-maintenance-parser/blob/develop/circuit_maintenance_parser/output.py#L14

chadell (commented):
I did a first review, ping me when ready for a second pass

nautobot_circuit_maintenance/jobs/site_search.py
jvanderaa: It's used for grouping of jobs.

nautobot_circuit_maintenance/jobs/site_search.py
jvanderaa: Sure, was keeping with the overflow issue found.

nautobot_circuit_maintenance/jobs/site_search.py
jvanderaa: Moved to overlap.

nautobot_circuit_maintenance/jobs/site_search.py
jvanderaa: Did a find and replace to move from ckt to circuit. `ckt` being short for circuit.

nautobot_circuit_maintenance/jobs/site_search.py
jvanderaa: Trying to keep the length of the line short. Added.

nautobot_circuit_maintenance/jobs/site_search.py
jvanderaa: What maintenance notifications would have no impact? 

I'd like to still be warned that there is overlap within the Job itself, even if there is no anticipated impact. For the nature of it. It is just a Circuit Maintenance Overlap check.

nautobot_circuit_maintenance/jobs/site_search.py
jvanderaa: Why keep it as debug versus success? I would like to see green checks to help see that there was a check and no overlap was found.

nautobot_circuit_maintenance/jobs/site_search.py
jvanderaa: ✅

nautobot_circuit_maintenance/jobs/site_search.py
jvanderaa: What do we mean by this?

nautobot_circuit_maintenance/jobs/site_search.py
chadell: My understanding of "success" is not about the network status health, but the health of the Job execution. So, from the Job point of view, this is just an info step that it has been processed and evaluated. Even when there are overlaps, the overlal Job should "success" and show the report with the issues.

nautobot_circuit_maintenance/jobs/site_search.py
chadell: as commented offline, let's leave this for ALL types, but add a note that about the impact type for future refinement if needed

nautobot_circuit_maintenance/jobs/site_search.py
chadell: Warning is good enough for overlappings

nautobot_circuit_maintenance/tests/test_jobs.py
chadell: sorry Josh, not following your point. I don't see why the MockRecord is necessary, it's exactly the same (from my point of view)

nautobot_circuit_maintenance/jobs/site_search.py
```diff
@@ -0,0 +1,159 @@
...
+
+    Args:
+        NamedTuple (NamedTupe): Parent class
+    """
+
+    start: datetime
+    end: datetime
+
+
+def check_for_overlap(record1: CircuitMaintenance, record2: CircuitMaintenance):
```
chadell, line 30: thinking alound, would make sense to generalize this function to `**records`, so you could find overlapping for more than only 2 maintenances? It works for the current case, and it will work for future cases where more than 2 maintenances could be taken into account, in one shot.
It's simply adding a for loop for ranges, and get the same max and min, no?

nautobot_circuit_maintenance/jobs/site_search.py
jvanderaa: Went with `log_info` instead of `log_success`. And will be adding a log_success at the bottom

nautobot_circuit_maintenance/jobs/site_search.py
jvanderaa: Maybe. In the longer run I think we need to compare the number of overlapping records to the number of circuits. If there are 4 circuits to a site, then as long as there are no more than 3 maintenances that overlap, we are good. May need some logic thinking on that one.

chadell: ok, let's leave for another iteration

nautobot_circuit_maintenance/__init__.py
```diff
@@ -90,6 +90,7 @@ class CircuitMaintenanceConfig(PluginConfig):
         "raw_notification_initial_days_since": 7,
         "raw_notification_size": 8192,
         "dashboard_n_days": 30,
+        "overlap_job_log_only_impacts": False,
```
chadell, line 93: we should add this to the Readme to explain it. And, personally, I would prefer a name like "overlap_job_exclude_no_impact_maintenances"

nautobot_circuit_maintenance/jobs/site_search.py
```diff
@@ -0,0 +1,176 @@
...
+    """Nautobot Job definition for finding sites without redundant circuit for impactful maintenance.
+
+    Args:
+        Job (Nautobot Job): Nautobot Job parent class
+    """
+
+    class Meta:
+        """Meta definition for the Job."""
+
+        name = "Find Sites With Circuit Overlap"
```
chadell, line 121:
```suggestion
        name = "Find Sites With Circuit Maintenance Overlap"
```

nautobot_circuit_maintenance/jobs/site_search.py
```diff
@@ -0,0 +1,176 @@
...
+        description = "Search for sites with multiple circuits, 1 or more circuit impacts."
```
chadell, line 122:
```suggestion
        description = "Search for sites with overlapping circuit maintenances, 1 or more circuit impacts, assuming only TWO circuits per site."
```

nautobot_circuit_maintenance/jobs/site_search.py
```diff
@@ -0,0 +1,176 @@
...
+            # Check both termination A and Z for values of None, add that record to the set for each site
+            for term in [circuit.termination_a, circuit.termination_z]:
+                if term is not None:
+                    return_dictionary[term.site.name].add(record)
+
+    return dict(return_dictionary)
+
+
+class FindSitesWithMaintenanceOverlap(Job):
+    """Nautobot Job definition for finding sites without redundant circuit for impactful maintenance.
```
chadell, line 112: I would add some notes about the design decisions/assumptions, such as 2 circuits per site, to enable future work.

nautobot_circuit_maintenance/jobs/site_search.py
```diff
@@ -0,0 +1,176 @@
...
+                                        message=f"There is an overlapping maintenance for site: {site.name}. Other maintenances: {other_circuit_maint}|{circuit_maint}",
+                                    )
+                        else:
+                            self.log_warning(
+                                obj=site,
+                                message=f"There is an overlapping maintenance for site: {site.name}. Other maintenances: {other_circuit_maint}|{circuit_maint}",
+                            )
+                    else:
+                        # Log success for each time there is a known circuit still available at the site at the same time
+                        self.log_info(
```
chadell, line 169: I still think this should be "debug", so you could easily exclude with the `debug` checkbox

chadell (commented):
some comments

README.md
```diff
@@ -33,6 +33,7 @@ PLUGINS_CONFIG = {
         "raw_notification_initial_days_since": 100,
         "raw_notification_size": 16384,
         "dashboard_n_days": 30,  # Defaults to 30 days in the configurations, change/override here
+        "overlap_job_log_only_impacts": False, # Defaults for False to alert on any notification, regardless of impact. True if there needs to be a known impact.
```
chadell, line 36: Sorry @jvanderaa for the confusion, I think I didn't explain me well.
Impacts are always known, as being a Pydantic class (in the parser), it always comes from a good place.
My point was to "alarm" or not, depending on the type of the IMPACT, leaving the `NO_IMPACT` circuits out of the check, because no issue is expected on them.

nautobot_circuit_maintenance/tests/test_jobs.py
chadell: ping

README.md
jvanderaa: What do we want to put here for a name. We are good, but I think we can update the var name appropriately.

nautobot_circuit_maintenance/jobs/site_search.py
```diff
@@ -0,0 +1,180 @@
...
+            # Check to see if there are any circuit maintenances that are overlapping at some moment in time
+            for site in site_list:
+                for other_circuit_maint in circuit_maintenance_mapper[site.name]:
+                    # Report failures for any time where a circuit will take an outage
+                    if circuit_maint == other_circuit_maint:
+                        continue
+                    if check_for_overlap(circuit_maint, other_circuit_maint):
+                        if PLUGIN_SETTINGS.get("overlap_job_log_only_impacts"):
+                            for impact in circuit_maint.circuitimpact_set:
+                                if impact.impact in CIRCUIT_IMPACT_SET:
```
chadell, line 163:
```suggestion
                                if impact.impact != NO-IMPACT":
```

nautobot_circuit_maintenance/tests/test_jobs.py
jvanderaa: Conversed and I think we are good with the new name.

docs/jobs.md
```diff
@@ -0,0 +1,40 @@
...
+- Ack: boolean to show if the maintenance has been properly handled by the operator.
+- Circuits: list of circuits and its specific impact linked to this maintenance.
+- Notes: list of internal notes linked to this maintenance.
+- Notifications: list of all the parsed notifications that have been processed for this maintenance.
+
+![Circuit Maintenance Job](images/circuit_maintenance.png)
+
+## Circuit Overlap Job
+
+The circuit overlap job that gets included with the Circuit Maintenance Plugin is a job that is going to search for possible overlapping maintenances, which **may** cause an outage of a site. The variable `overlap_job_log_only_impacts` controls on the check if a maintenance notification has an expected impact. Default is `False` for this setting, that any maintenance notification will be alerted on within the Nautobot Job.
```
chadell, line 38:
```suggestion
The circuit overlap job that gets included with the Circuit Maintenance Plugin is a job that is going to search for possible overlapping maintenances, which **may** cause an outage of a site. The variable `overlap_job_exclude_no_impact ` controls on the check if a maintenance notification has an expected impact. Default is `False` for this setting, that any maintenance notification will be alerted on within the Nautobot Job. 
```

nautobot_circuit_maintenance/__init__.py
chadell, line 93:
```suggestion
        "overlap_job_exclude_no_impact": False,
```

chadell (approved):
LGTM

Issue #210, opened 2022-09-21
Dockerfile results in unexpected Nautobot version being installed


glennmatthews:
### Environment

CI, local dev environment

### Expected Behavior

When building with a specified `NAUTOBOT_VER`, the built Docker image should have that Nautobot version installed.

### Observed Behavior

Unexpected version is installed, for example in https://github.com/nautobot/nautobot-plugin-circuit-maintenance/actions/runs/3050285818/jobs/4917347435, although the build starts from image `ghcr.io/nautobot/nautobot-dev:1.3.8-py3.7`, the final image ends up with Nautobot 1.4.0:

```
Step 16/17 : RUN poetry install --no-interaction --no-ansi -E metrics
 ---> Running in a32f99d0b949
Skipping virtualenv creation, as specified in config file.
Installing dependencies from lock file

Package operations: 1 install, 7 updates, 0 removals

  • Updating importlib-metadata (4.4.0 -> 4.12.0)
  • Updating jsonschema (4.4.0 -> 4.7.2)
  • Updating django-cors-headers (3.10.1 -> 3.13.0)
  • Updating django-extensions (3.1.5 -> 3.2.0)
  • Installing django-tree-queries (0.11.0)
  • Updating drf-yasg (1.20.3 -> 1.21.3)
  • Updating pillow (9.0.1 -> 9.2.0)
  • Updating nautobot (1.3.8 -> 1.4.0)    <---------------------<<<

Installing the current project: nautobot-circuit-maintenance (0.6.0)
Removing intermediate container a32f99d0b949
```

### Steps to Reproduce

`INVOKE_NAUTOBOT_CIRCUIT_MAINTENANCE_NAUTOBOT_VER=1.4.3 invoke build`, for example, starts from the Nautobot 1.4.3 base image but ends up with Nautobot 1.4.0 installed as well.

Comparing this project's Dockerfile against https://github.com/nautobot/nautobot-plugin-firewall-models/blob/develop/development/Dockerfile (which does *not* show this issue) may prove helpful.

chadell:
Fixed by #215

PR #215, opened 2022-09-27
Fix Dockerfile to take NAUTBOT_VER into account when installing dependencies

chadell:
Addresses #210

### DIFF
--- a/development/Dockerfile
+++ b/development/Dockerfile
@@ -19,11 +19,10 @@ RUN poetry config virtualenvs.create false \
 # -------------------------------------------------------------------------------------
 # Install Nautobot Plugin
 # -------------------------------------------------------------------------------------
-
-WORKDIR /source
+WORKDIR /tmp/install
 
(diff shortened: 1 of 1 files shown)


glennmatthews:
Doesn't look like a complete fix - https://github.com/nautobot/nautobot-plugin-circuit-maintenance/actions/runs/3135597953/jobs/5091713412 still ends up with Nautobot 1.4.0. I think at least part of the issue is [here](https://github.com/nautobot/nautobot-plugin-circuit-maintenance/blob/develop/development/Dockerfile#L38), as `COPY . /source` is overwriting the updated `pyproject.toml` and `poetry.lock` with the originals from Git.

Comparing to https://github.com/nautobot/nautobot-plugin-firewall-models/blob/develop/development/Dockerfile, there are a couple of extra steps involving doing the `poetry add nautobot` in a `/tmp/install` directory and then copying from there to `/source` *after* `COPY . /source` so that the updated files take precedence.

chadell:
you are right, I missed that point

glennmatthews (approved):

PR #219, opened 2022-10-10
Limit working nautobot version 1.3.9

chadell:


chadell:
will be fixed by https://github.com/nautobot/nautobot/pull/2602

PR #221, opened 2022-10-10
Fixes where duplicate maintence records show jobs

### LINKED ISSUE #220: Overlap Search Shows Multiple Results with Multiple Notification Sources
### Environment

- Python version:3.x
- Nautobot version: 1.3.9
- nautobot-circuit-maintenance version: develop

### Expected Behavior

Expected one occurrence of overlaps

### Observed Behavior

Found multiple occurrences

### Steps to Reproduce

1. Create multiple sources of notifications
2. Run the job to check Overlap
3. Find multiple instances

jvanderaa:
Fixes #220 

* Adds a list of maintenance records that have been processed
* If the maintenance that is being checked is in the maintenance list, stop checking it

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v0.7.0
+
+- Adds ability to search for overlapping jobs
+
 ## v0.6.0
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,12 @@
 
 ## v0.6.2
 
+### Fixed
+
 - #221 - Fixes error in Overlap Job
+
+### Added
+
 - #221 - Adds watchmedo to auto reload worker containers during job development
 
 ## v0.6.1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,8 @@
 
 ## v0.6.2
 
-- #221 Fixes error in Overlap Job
-- #221 Adds watchmedo to auto reload worker containers during job development
+- #221 - Fixes error in Overlap Job
+- #221 - Adds watchmedo to auto reload worker containers during job development
 
 ## v0.6.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,18 @@
 # Changelog
 
+## v0.6.0
+
+### Added
+
+- #191 - Add Dashboard View
+- #187 - Add maintenance start date to Circuit Maintenance extension in Device detail view
+
+### Changed
+
+- #195 - Drop Python 3.6 support
+- #182 - Add Readme reference to post_upgrade
+- #180 - Update dependency google-auth-oauthlib to ^0.5.0
+
 ## v0.5.0
 
 ### Added
@@ -9,7 +22,6 @@
 ### Changed
 
 - #165 - Numerous UI/UX refinements and improvements. See #176 for details.
-
 
 ## v0.4.3
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,8 @@
 
 ## v0.6.2
 
-- (#221) Fixes error in Overlap Job
-- (#221) Adds watchmedo to auto reload worker containers during job development
+- #221 Fixes error in Overlap Job
+- #221 Adds watchmedo to auto reload worker containers during job development
 
 ## v0.6.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,21 @@
 # Changelog
 
-## v0.7.0
+## v0.6.2
 
-- Adds ability to search for overlapping jobs
+- (#221) Fixes error in Overlap Job
+- (#221) Adds watchmedo to auto reload worker containers during job development
+
+## v0.6.1
+
+### Added
+
+- #214 - Maintenance Overlap Job Addition
+
+### Fixed
+
+- #222 - Fix for circuit maintenance detail view. Show approriate A and Z termination sides.
+- #216 - Fix variable naming in Overlapping Maintenance Job: start_date instead of start_time.
+- #208 - Fix CI credentials.
 
 ## v0.6.0
 
--- a/development/Dockerfile
+++ b/development/Dockerfile
@@ -1,8 +1,6 @@
-ARG NAUTOBOT_VER="1.3.5"
+ARG NAUTOBOT_VER="1.4.1"
 ARG PYTHON_VER=3.8
 FROM ghcr.io/nautobot/nautobot-dev:${NAUTOBOT_VER}-py${PYTHON_VER}
-
-ARG NAUTOBOT_VER=${NAUTOBOT_VER}
 
 ENV prometheus_multiproc_dir=/prom_cache
 
@@ -21,13 +19,17 @@ RUN poetry config virtualenvs.create false \
 # -------------------------------------------------------------------------------------
 # Install Nautobot Plugin
 # -------------------------------------------------------------------------------------
-
-WORKDIR /source
+WORKDIR /tmp/install
 
 # Copy in only pyproject.toml/poetry.lock to help with caching this layer if no updates to dependencies
-COPY poetry.lock pyproject.toml /source/
+COPY poetry.lock pyproject.toml /tmp/install/
 
-# Add Nautobot to the poetry file for testing
+# Add the requested Nautobot version to pyproject
+# to install the correct version based on the NAUTOBOT_VER argument
+# Otherwise Poetry will override the version in this container
+# with the one in the poetry.lock
+# Redifine NAUTOBOT_VER as a build arg as initial is defined outside FROM.
+ARG NAUTOBOT_VER="1.4.1"
 RUN poetry add nautobot==$NAUTOBOT_VER
 
 # --no-root declares not to install the project package since we're wanting to take advantage of caching dependency installation
@@ -35,7 +37,9 @@ RUN poetry add nautobot==$NAUTOBOT_VER
 RUN poetry install --no-interaction --no-ansi --no-root -E metrics
 
 # Copy in the rest of the source code and install local Nautobot plugin
+WORKDIR /source
 COPY . /source
+RUN cp /tmp/install/* /source/
 RUN poetry install --no-interaction --no-ansi -E metrics
 
 COPY development/nautobot_config.py /opt/nautobot/nautobot_config.py
--- a/development/Dockerfile
+++ b/development/Dockerfile
@@ -1,4 +1,4 @@
-ARG NAUTOBOT_VER="1.4.1"
+ARG NAUTOBOT_VER="1.4.7"
 ARG PYTHON_VER=3.8
 FROM ghcr.io/nautobot/nautobot-dev:${NAUTOBOT_VER}-py${PYTHON_VER}
 
@@ -29,7 +29,7 @@ COPY poetry.lock pyproject.toml /tmp/install/
 # Otherwise Poetry will override the version in this container
 # with the one in the poetry.lock
 # Redifine NAUTOBOT_VER as a build arg as initial is defined outside FROM.
-ARG NAUTOBOT_VER="1.4.1"
+ARG NAUTOBOT_VER="1.4.7"
 RUN poetry add nautobot==$NAUTOBOT_VER
 
 # --no-root declares not to install the project package since we're wanting to take advantage of caching dependency installation
--- a/nautobot_circuit_maintenance/jobs/site_search.py
+++ b/nautobot_circuit_maintenance/jobs/site_search.py
@@ -101,7 +101,7 @@ def build_sites_to_maintenance_mapper(maintenance_queryset):
         for circuit in record.circuits:
             # Check both termination A and Z for values of None, add that record to the set for each site
             for term in [circuit.termination_a, circuit.termination_z]:
-                if term is not None:
+                if term.site is not None:
                     return_dictionary[term.site.name].add(record)
 
     return dict(return_dictionary)
@@ -167,12 +167,12 @@ class FindSitesWithMaintenanceOverlap(Job):
                                 if impact.impact != "NO-IMPACT":
                                     self.log_warning(
                                         obj=site,
-                                        message=f"There is an overlapping maintenance for site: {site.name} on {circuit_maint.start_date}. Other maintenances: {other_circuit_maint}|{circuit_maint}",
+                                        message=f"There is an overlapping maintenance for site: {site.name} on {circuit_maint.start_time}. Other maintenances: {other_circuit_maint}|{circuit_maint}",
                                     )
                         else:
                             self.log_warning(
                                 obj=site,
-                                message=f"There is an overlapping maintenance for site: {site.name} on {circuit_maint.start_date}. Other maintenances: {other_circuit_maint}|{circuit_maint}",
+                                message=f"There is an overlapping maintenance for site: {site.name} on {circuit_maint.start_time}. Other maintenances: {other_circuit_maint}|{circuit_maint}",
                             )
                     else:
                         # Log success for each time there is a known circuit still available at the site at the same time
--- a/nautobot_circuit_maintenance/jobs/site_search.py
+++ b/nautobot_circuit_maintenance/jobs/site_search.py
@@ -1,7 +1,6 @@
 """Site searching Job definition."""
 import collections
-from datetime import datetime, date
-from typing import NamedTuple
+from datetime import date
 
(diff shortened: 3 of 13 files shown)


nautobot_circuit_maintenance/jobs/site_search.py
```diff
@@ -143,6 +143,7 @@ def run(self, data=None, commit=None):
         # Loop over each of the circuit maintenance records
         # pylint: disable=too-many-nested-blocks
         counter: int = 0
+        maintenance_list = []
```
chadell, line 146: making this list "global", for all sites, means that if a maintenance affects circuits in multiple sites, it's only taken into account on the first site.

chadell: the name `maintenance_list` is not super meaningful...

nautobot_circuit_maintenance/jobs/site_search.py
```diff
@@ -154,6 +155,10 @@ def run(self, data=None, commit=None):
             for site in site_list:
                 for other_circuit_maint in circuit_maintenance_mapper[site.name]:
                     # Report failures for any time where a circuit will take an outage
+                    maintenance_list.append(circuit_maint)
```
chadell, line 158: why we try to append  the `circuit_maint` from the outer `circuit_maintenances` loop, within the `site_list` loop. It looks counterintuitive to me. I would di is as soon as possible.
Also, seeing this, the `list` maybe better a `set`?

chadell (changes requested):
some comments on the logic

nautobot_circuit_maintenance/jobs/site_search.py
jvanderaa: Updated the var name. 

Thinking about the logic, it's more that overlap has been checked for in general. Just a list of maintenances, not specific for a site.

nautobot_circuit_maintenance/jobs/site_search.py
jvanderaa: I had some issues with sets, I like sets. But the list was able to handle it.

nautobot_circuit_maintenance/jobs/site_search.py
jvanderaa: I did dive down in on this. If we move the `processed_circuit_maintenances` down one loop, this causes an issue. There are too many overlaps found.

Can we pair on this in my morning? Perhaps 8:30am ET? @chadell

jvanderaa:
@gsnider2195 thank you today, this is the PR that is being worked through.

nautobot_circuit_maintenance/jobs/site_search.py
```diff
@@ -37,22 +25,13 @@ def check_for_overlap(record1: CircuitMaintenance, record2: CircuitMaintenance):
     Returns:
         bool: Result of there is overlap
```
gsnider2195, line 26:
```suggestion
        bool: True if there is overlap, otherwise False.
```

nautobot_circuit_maintenance/jobs/site_search.py
```diff
@@ -35,24 +23,15 @@ def check_for_overlap(record1: CircuitMaintenance, record2: CircuitMaintenance):
...
-    range1 = DateTimeRange(start=record1.start_time, end=record1.end_time)
-    range2 = DateTimeRange(start=record2.start_time, end=record2.end_time)
-
-    # Determine the latest start and the earliest end to determine overlap
-    latest_start = max(range1.start, range2.start)
-    earliest_end = min(range1.end, range2.end)
-
-    # Get the delta, compare versus 0 (negative number is no overlap)
-    delta = (earliest_end - latest_start).days + 1
+    if record1.end_time < record2.start_time or record1.start_time > record2.end_time:
```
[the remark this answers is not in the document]
gsnider2195: Should this allow maintenances that start and end at the same time? For example Maintenance A ends at 8pm and Maintenance B starts at 8pm? If so, these need to be changed to `<=`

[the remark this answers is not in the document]
jvanderaa: We should be good with the set up, the end times are more of a fuzzy end when working with carriers. It is more that is the full window, and all thing should be wrapped up sooner.

jvanderaa:
@chadell anything further?

development/Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG NAUTOBOT_VER="1.4.1"
+ARG NAUTOBOT_VER="1.4.7"
```
chadell, line 1: Is this some requirement? or just to update it?

nautobot_circuit_maintenance/jobs/site_search.py
```diff
@@ -101,10 +80,9 @@ def build_sites_to_maintenance_mapper(maintenance_queryset):
         for circuit in record.circuits:
             # Check both termination A and Z for values of None, add that record to the set for each site
             for term in [circuit.termination_a, circuit.termination_z]:
-                # If the circuit is connected to a provider network, then it will return None for a site.
-                # Also only checks connected circuits.
-                if term.site is not None:
-                    return_dictionary[term.site.name].add(record)
+                if term is not None:
+                    if term.site is not None:
```
chadell, line 84:
```suggestion
                if term is not None and term.site is not None:
```

nautobot_circuit_maintenance/jobs/site_search.py
```diff
@@ -144,38 +124,45 @@ def run(self, data=None, commit=None):
...
                         else:
                             self.log_warning(
                                 obj=site,
                                 message=f"There is an overlapping maintenance for site: {site.name} on {circuit_maint.start_time}. Other maintenances: {other_circuit_maint}|{circuit_maint}",
                             )
-                    else:
-                        # Log success for each time there is a known circuit still available at the site at the same time
-                        self.log_debug(message="Checked maintenance for overlap, no overlap was found.")
+
+            # Log success for each time there is a known circuit still available at the site at the same time
```
chadell, line 160: this is "not" accurate. Maybe the site has 3 circuits, but we only check for overlap, assuming only 2 circuits per site. As we expect to improve this in the future, I would make it more clear what we do, so when we try to improve it, we will easily understand the logic

nautobot_circuit_maintenance/jobs/site_search.py
```diff
@@ -144,38 +124,45 @@ def run(self, data=None, commit=None):
...
+            if overlapping_maintenance is False:
+                if data["job_debug"]:
+                    self.log_info(obj=circuit_maint, message="Checked maintenance for overlap, no overlap was found.")
 
         self.log_success(
             obj=None,
-            message=f"Successfully checked through {counter} maintenance notification{'s' if counter > 1 else ''}.",
+            message=f"Successfully checked through {circuit_maintenances.count()} maintenance notification{'s'[:circuit_maintenances.count()^1]}.",
```
chadell, line 167: sorry, but I don't get `^1`, I tried it and I don't understand the logic

nautobot_circuit_maintenance/jobs/site_search.py
```diff
@@ -144,38 +124,45 @@ def run(self, data=None, commit=None):
...
                         continue
                     if check_for_overlap(circuit_maint, other_circuit_maint):
+                        overlapping_maintenance = True
                         if PLUGIN_SETTINGS.get("overlap_job_exclude_no_impact"):
                             for impact in circuit_maint.circuitimpact_set:
                                 if impact.impact != "NO-IMPACT":
                                     self.log_warning(
                                         obj=site,
-                                        message=f"There is an overlapping maintenance for site: {site.name} on {circuit_maint.start_time}. Other maintenances: {other_circuit_maint}|{circuit_maint}",
+                                        message=f"There is an overlapping __no-impact__ maintenance for site: {site.name} on {circuit_maint.start_time}. Other maintenances: {other_circuit_maint}|{circuit_maint}",
```
chadell, line 152: I believe that this is wrong. if we are here, this impact is NOT (!=) "NO-IMPACT", so it's a regular overlapping. Maybe it would be good to add a DEBUG log, for the "NO-IMPACT", saying that there is overlap, but becuase it has no impact, we simply log it, without complaining.

nautobot_circuit_maintenance/jobs/site_search.py
```diff
@@ -144,38 +124,45 @@ def run(self, data=None, commit=None):
...
                 for other_circuit_maint in circuit_maintenance_mapper[site.name]:
                     # Report failures for any time where a circuit will take an outage
-                    if circuit_maint == other_circuit_maint:
+                    # Check to see if the other circuit maintenance has already had overlap checked.
+                    # If this is the same circuit maintenance record, that will also show up in the processed list
+                    # since the list is being added to before the loop.
+                    if other_circuit_maint in processed_maintenance_list:
                         continue
                     if check_for_overlap(circuit_maint, other_circuit_maint):
+                        overlapping_maintenance = True
```
chadell, line 146: thinking on this, I believe that we should only make this `True`, when we find a dangerous overlapping. I mean, if we have the `overlap_job_exclude_no_impact` flag True, and all the impact are of this type, the overlapping is not relevant to me, so we should not mark it True, and take it into account.

chadell (changes requested):

development/Dockerfile
jvanderaa: Just to update to a version that works for the bug found in Nautobot

nautobot_circuit_maintenance/jobs/site_search.py
jvanderaa: Bit wise "magic". @gsnider2195

nautobot_circuit_maintenance/jobs/site_search.py
jvanderaa: This gets back into the discussion about `NO-IMPACT` notifications and the setting that is in place to address this. In my experience I would always want to warn about any notifications when it came to circuits, even if one is sent out for `NO-IMPACT`. It's like a change request that has no **expected** impact. But there is risk because something is being done.

nautobot_circuit_maintenance/jobs/site_search.py
jvanderaa: I think that I have this in place now, that it should follow along.

nautobot_circuit_maintenance/jobs/site_search.py
jvanderaa: Basically if the count is 1 the result is 0 which will return nothing in the string. If there is any other integer count, then it returns the rest of the string.

gsnider2195: It's XORing the count. 1 returns 0, everything else returns 1. Then it's using that to slice the string. So:

```
checked through 0 maintenance notifications.
checked through 1 maintenance notification.
checked through 2 maintenance notifications.
```

gsnider2195: > Basically if the count is 1 the result is 0 which will return nothing in the string. If there is any other integer count, then it returns the rest of the string.

I didn't see this reply. This is the answer 😄

chadell (changes requested):

jvanderaa:
Take a peak on this commit message! @chadell

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## v0.6.2
+
+- (#221) Fixes error in Overlap Job
```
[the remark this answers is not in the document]
jvanderaa: Following the format that was put in place on pynautobot.... https://github.com/nautobot/pynautobot/blob/develop/CHANGELOG.md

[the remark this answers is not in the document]
jvanderaa: Updating here, I knew that I got it from somewhere, and it was pynautobo.t

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## v0.6.2
+
+- #221 Fixes error in Overlap Job
+- #221 Adds watchmedo to auto reload worker containers during job development
```
chadell, line 6:
```suggestion
- #221 - Fixes error in Overlap Job
- #221 - Adds watchmedo to auto reload worker containers during job development
```

nautobot_circuit_maintenance/jobs/site_search.py
```diff
@@ -166,16 +149,21 @@ def run(self, data=None, commit=None):
...
                             )
-                    else:
-                        # Log success for each time there is a known circuit still available at the site at the same time
-                        self.log_debug(message="Checked maintenance for overlap, no overlap was found.")
+                            overlapping_maintenance = True
+
+            # Log success for when there is not an overlapping maintenance at a site
+            if overlapping_maintenance is False:
+                if data["job_debug"]:
+                    self.log_info(obj=circuit_maint, message="Checked maintenance for overlap, no overlap was found.")
```
chadell, line 164:
```suggestion
            if overlapping_maintenance is False and data["job_debug"]:
                self.log_info(obj=circuit_maint, message="Checked maintenance for overlap, no overlap was found.")
```

nautobot_circuit_maintenance/jobs/site_search.py
```diff
@@ -166,16 +149,21 @@ def run(self, data=None, commit=None):
                                         obj=site,
                                         message=f"There is an overlapping maintenance for site: {site.name} on {circuit_maint.start_time}. Other maintenances: {other_circuit_maint}|{circuit_maint}",
                                     )
+                                overlapping_maintenance = True
```
chadell, line 152: I would change to something like:

```python
                    if check_for_overlap(circuit_maint, other_circuit_maint):
                        overlapping_maintenance = True
                        message = f"There is an overlapping maintenance for site: {site.name} on {circuit_maint.start_time}. Other maintenances: {other_circuit_maint}|{circuit_maint}"
                        if PLUGIN_SETTINGS.get("overlap_job_exclude_no_impact"):
                            for impact in circuit_maint.circuitimpact_set:
                                if impact.impact != "NO-IMPACT":
                                    break
                            else:
                                 continue
                                    
                        self.log_warning(
                            obj=site,
                            message=message,
                        )
```

chadell (commented):
some "optimization" comment, have a look

nautobot_circuit_maintenance/jobs/site_search.py
```diff
@@ -144,38 +123,43 @@ def run(self, data=None, commit=None):
...
-                                message=f"There is an overlapping maintenance for site: {site.name} on {circuit_maint.start_time}. Other maintenances: {other_circuit_maint}|{circuit_maint}",
-                            )
-                    else:
-                        # Log success for each time there is a known circuit still available at the site at the same time
-                        self.log_debug(message="Checked maintenance for overlap, no overlap was found.")
+                            for circuit_impact in circuit_maint.circuitimpact_set:
+                                if circuit_impact.impact != "NO-IMPACT":
+                                    break
+
+                                continue
```
chadell, line 150:
```suggestion
                          else:
                                continue
```

the else is used with the `for`. If no iteration of the loop hits break (so all the impacts are NO-IMPACT), we do a `continue` to skip logging a warning message

nautobot_circuit_maintenance/jobs/site_search.py
```diff
@@ -144,38 +123,43 @@ def run(self, data=None, commit=None):
...
+
+                        self.log_warning(
+                            obj=site,
+                            message=f"There is an overlapping maintenance for site: {site.name} on {circuit_maint.start_time}. Other maintenances: {other_circuit_maint}|{circuit_maint}",
+                        )
+                        overlapping_maintenance = True
```
chadell, line 156: If I understand correctly, `overlapping_maintenance = True` should be updated right after the `if check_for_overlap(circuit_maint, other_circuit_maint):`.

Why? because even in the case of the option for no-impact, and having all the circuit of this type, you are detecting and `overlapping_maintenance`, but not logging

nautobot_circuit_maintenance/jobs/site_search.py
jvanderaa: That confuses me.

jvanderaa: `R1723: Unnecessary "else" after "break", remove the "else" and de-indent the code inside it (no-else-break)`

chadell: I'm doing wrong them.
The `else` should be indented to the `for` level,  it's not for the `if`. I could try locally

jvanderaa: Let me know @chadell

chadell (approved):
LGTM

scetron (approved):
LGTM

PR #228, opened 2022-12-06
Enable Upstream Testing

bryanculver:
- Updated Dockerfile to be compatible with upstream testing
- Added workflow fragment to trigger testing
- Updated CI to run 3.10 unittests
- Bumped Nautobot minimal version to 1.4
- Fixed tests failing with Nautobot 1.5.12
- Disabled tests that can't be fixed, blocked by https://github.com/nautobot/nautobot/issues/3419

### DIFF
--- a/.github/workflows/upstream_testing.yml
+++ b/.github/workflows/upstream_testing.yml
@@ -4,8 +4,6 @@ name: "Nautobot Upstream Monitor"
 on:  # yamllint disable-line rule:truthy rule:comments
   schedule:
     - cron: "0 4 */2 * *"  # every other day at midnight
-  push:  # yamllint disable rule:empty-values
-# TODO: REMOVE PUSH LINE ABOVE
 
 jobs:
   upstream-test:
--- a/.github/workflows/upstream_testing.yml
+++ b/.github/workflows/upstream_testing.yml
@@ -1,7 +1,7 @@
 ---
-name: "Nautobot Upstream Testing"
+name: "Nautobot Upstream Monitor"
 
-on: # yamllint disable-line rule:truthy rule:comments
+on:  # yamllint disable-line rule:truthy rule:comments
   schedule:
     - cron: "0 4 */2 * *"  # every other day at midnight
   push:  # yamllint disable rule:empty-values
--- a/.github/workflows/upstream_testing.yml
+++ b/.github/workflows/upstream_testing.yml
@@ -1,0 +1,15 @@
+---
+name: "Nautobot Upstream Testing"
+
+on: # yamllint disable-line rule:truthy rule:comments
+  schedule:
+    - cron: "0 4 */2 * *"  # every other day at midnight
+  push:  # yamllint disable rule:empty-values
+# TODO: REMOVE PUSH LINE ABOVE
+
+jobs:
+  upstream-test:
+    uses: "nautobot/nautobot/.github/workflows/plugin_upstream_testing_base.yml@develop"
+    with:  # Below could potentially be collapsed into a single argument if a concrete relationship between both is enforced
+      invoke_context_name: "NAUTOBOT_CIRCUIT_MAINTENANCE"
+      plugin_name: "nautobot_circuit_maintenance"
--- a/development/Dockerfile
+++ b/development/Dockerfile
@@ -1,13 +1,13 @@
 # -------------------------------------------------------------------------------------
-# Nautobot Plugin Development Dockerfile Template
-# Version: 1.0.0
+# Nautobot App Developement Dockerfile Template
+# Version: 1.1.0
 #
-# Plugins that need to add additional steps or packages can do in the section below.
+# Apps that need to add additional steps or packages can do in the section below.
 # -------------------------------------------------------------------------------------
 # !!! USE CAUTION WHEN MODIFYING LINES BELOW
 
-# Accepts a desired Nautobot version as build argument, default to 1.4.7
-ARG NAUTOBOT_VER="1.4.7"
+# Accepts a desired Nautobot version as build argument, default to 1.4.10
+ARG NAUTOBOT_VER="1.4.10"
 
 # Accepts a desired Python version as build argument, default to 3.8
 ARG PYTHON_VER="3.8"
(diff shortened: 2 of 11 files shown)


development/Dockerfile
```diff
@@ -1,45 +1,73 @@
...
+#  so pip can use it to recognize local constraints.
+RUN pip show nautobot | grep "^Version: " | sed -e 's/Version: /nautobot==/' > constraints.txt
+
+# Use Poetry to grab dev dependencies from the lock file
+# Can be improved in Poetry 1.2 which allows `poetry install --only dev`
+#
+# We can't use the entire freeze as it takes forever to resolve with rigidly fixed non-direct dependencies,
+#   especially those that are only direct to Nautobot but the container included versions slightly mismatch
+RUN poetry export -f requirements.txt --without-hashes --output poetry_freeze_base.txt
+RUN poetry export -f requirements.txt --dev --extras metrics  --without-hashes --output poetry_freeze_all.txt
```
bryanculver, line 61: This deviates from the standard because this plugin publishes extras.

`--extras metrics` is what has been added

.github/workflows/upstream_testing.yml
```diff
@@ -0,0 +1,15 @@
+---
+name: "Nautobot Upstream Testing"
+
+on: # yamllint disable-line rule:truthy rule:comments
+  schedule:
+    - cron: "0 4 */2 * *"  # every other day at midnight
+  push:  # yamllint disable rule:empty-values
+# TODO: REMOVE PUSH LINE ABOVE
```
bryanculver, line 8:
```suggestion
```
Remove before merging.

bryanculver:
`next` is expected to fail because it goes to a new major version of Nautobot so can be safely ignored for now.

development/Dockerfile
```diff
@@ -1,45 +1,73 @@
...
+# !!! USE CAUTION WHEN MODIFYING LINES BELOW
+
+# Accepts a desired Nautobot version as build argument, default to 1.4.7
 ARG NAUTOBOT_VER="1.4.7"
-ARG PYTHON_VER=3.8
-FROM ghcr.io/nautobot/nautobot-dev:${NAUTOBOT_VER}-py${PYTHON_VER}
 
-ENV prometheus_multiproc_dir=/prom_cache
+# Accepts a desired Python version as build argument, default to 3.7
+ARG PYTHON_VER="3.8"
```
bryanculver, line 13: There is a mismatch here (and in the previous version of this file):

**Nautobot Version**
- `1.4.7` previous default in Dockerfile
- `latest` current invoke default: https://github.com/nautobot/nautobot-plugin-circuit-maintenance/blob/5bbf9c8763b9f6e255703ca269ba95b2c63df32c/tasks.py#L43
- `1.0.0-beta.4` current minimum in `PluginConfig`: https://github.com/nautobot/nautobot-plugin-circuit-maintenance/blob/5bbf9c8763b9f6e255703ca269ba95b2c63df32c/nautobot_circuit_maintenance/__init__.py#L86 
- `*` current minimum in `pyproject.toml`. https://github.com/nautobot/nautobot-plugin-circuit-maintenance/blob/5bbf9c8763b9f6e255703ca269ba95b2c63df32c/pyproject.toml#L23

**Python Version**
- `3.8` previous default in Dockerfile
- `3.7` current defaults in invoke and `pyproject.toml`

I will go ahead and update these in my PR but I would like to know the maintainer's preferences.

chadell: I would use Nautobot latest for development, while leaving the all Nautobot versions supported. For Python, I would go with 3.8 for development, and keep testing for 3.7 in the pipeline.
is this in line with your recommendations?

development/Dockerfile
```diff
@@ -1,45 +1,73 @@
...
-RUN poetry install --no-interaction --no-ansi -E metrics
 
-COPY development/nautobot_config.py /opt/nautobot/nautobot_config.py
+# Get container's installed Nautobot version as a forced constraint
+# NAUTOBOT_VER may be a branch name and not a published release therefor we need to get the installed version
+#  so pip can use it to recognize local constraints.
+RUN pip show nautobot | grep "^Version: " | sed -e 's/Version: /nautobot==/' > constraints.txt
+
+# Use Poetry to grab dev dependencies from the lock file
+# Can be improved in Poetry 1.2 which allows `poetry install --only dev`
```
chadell, line 56: why not using poetry 1.2 then?

development/Dockerfile
chadell: to better better understand. Why we don't use poetry to install instead of translating to requirements and install with Pip? (maybe it's somewhere explained by I missed it)

development/Dockerfile
glennmatthews: Older Nautobot-dev Docker images come with poetry 1.1.x. IIRC we'd have to explicitly upgrade poetry to latest as part of this Dockerfile, though that might be desirable in any case?

bryanculver: Glenn's point is correct. Is this blocking getting started with automated testing?

chadell: nope, just informational question

development/Dockerfile
chadell: ok, understood

chadell:
to refresh, @bryanculver what is missing in this PR to move it forward? how could we help you?

bryanculver:
> to refresh, @bryanculver what is missing in this PR to move it forward? how could we help you?

Making updates similar to those done in Chatops: https://github.com/nautobot/nautobot-plugin-chatops/pull/171

Namely:

- Bring up the minimum Nautobot version to 1.4+
- Update CI to ensure we're testing on the right versions
- Update to the latest example of the upstream_testing.yml and Dockerfiles (see cookie-cutter or Chatops PR for the example)
- Addressing conflicts like this: https://github.com/nautobot/nautobot-plugin-circuit-maintenance/pull/228#issuecomment-1339813292
    - These plugins are never intended to be installed standalone and should be running on dev base images so you should get dependencies like this for free

snaselj:
Rebased on top of `develop`

snaselj:
Rebased on top of `develop`

snaselj:
The last test failing is `upstream-test` on `next`

snaselj:
This PR is ready for review now.

chadell (approved):
🚀 LGTM!

Issue #231, opened 2022-12-27
Legacy templates need to extend `generic/object_detail.html`


HanlinMiao:
https://github.com/nautobot/nautobot-plugin-circuit-maintenance/blob/develop/nautobot_circuit_maintenance/templates/nautobot_circuit_maintenance/circuitimpact.html (Is this supposed to extend object_detail.html?)
https://github.com/nautobot/nautobot-plugin-circuit-maintenance/blob/develop/nautobot_circuit_maintenance/templates/nautobot_circuit_maintenance/circuitimpact.html (Is this supposed to extend object_detail.html?)
https://github.com/nautobot/nautobot-plugin-circuit-maintenance/blob/develop/nautobot_circuit_maintenance/templates/nautobot_circuit_maintenance/note.html (Is this supposed to extend object_detail.html?)
https://github.com/nautobot/nautobot-plugin-circuit-maintenance/blob/develop/nautobot_circuit_maintenance/templates/nautobot_circuit_maintenance/notificationsource.html (Is this supposed to extend object_detail.html?)
https://github.com/nautobot/nautobot-plugin-circuit-maintenance/blob/develop/nautobot_circuit_maintenance/templates/nautobot_circuit_maintenance/parsednotification.html (Is this supposed to extend object_detail.html?)
https://github.com/nautobot/nautobot-plugin-circuit-maintenance/blob/develop/nautobot_circuit_maintenance/templates/nautobot_circuit_maintenance/rawnotification.html (Is this supposed to extend object_detail.html?)

PR #232, opened 2022-12-31
Changed extends 'base.html' to extends 'generic/object_detail.html' for detail templates

briddo:
Fixes: #231


.idea/.gitignore
```diff
@@ -0,0 +1,8 @@
+# Default ignored files
```
glennmatthews, line 1: Need to add `.idea` to the project's base `.gitignore` and remove all of these files from Git.

nautobot_circuit_maintenance/templates/nautobot_circuit_maintenance/circuitimpact.html
```diff
@@ -1,4 +1,4 @@
-{% extends 'base.html' %}
+{% extends 'generic/object_detail.html' %}
```
glennmatthews, line 1: `object_detail.html` was only added in Nautobot 1.2 so this change will need to be accompanied by an increase in the project's required Nautobot min_version.

glennmatthews (changes requested):

nautobot_circuit_maintenance/templates/nautobot_circuit_maintenance/circuitimpact.html
glennmatthews, line 1: In addition to this change, need to review whether any or all of the contents of `block title` and` `block header` are now redundant, and if they can be removed to simplify the template.

Additionally, `block content` should be updated to only override `block content_left_page` as appropriate.

Same comment on the other template files below.

glennmatthews (changes requested):

glennmatthews:
Looks like there's something incorrect in the `parsednotification.html` that's causing template rendering to error out in the tests.

gsnider2195:
> Looks like there's something incorrect in the `parsednotification.html` that's causing template rendering to error out in the tests.

There was a bug in 1.2 that was fixed in https://github.com/nautobot/nautobot/pull/1366 that we're running into here because `ParsedNotification` doesn't have a list view. We'll have to override the header block in that file or create a list view for `ParsedNotification`.

glennmatthews (approved):

chadell:
@glennmatthews have you verified to rendered outcome? if so, it LGTM

glennmatthews:
Locally the only data I have is some configured Notification Sources; that detail view looks fine to me. I don't have data for the other models so I can't verify them manually at this time. The diffs *look* good to me though and the tests are passing so that would catch any egregious template-rendering errors.

PR #246, opened 2023-03-22
Fix RTD build errors

### LINKED ISSUE #245: Docs: RTD build failing
### Environment

- `develop` branch docs (i.e. latest)

### Expected Behavior

- Hosted docs on readthedocs.org should build successfully.

### Observed Behavior

- Build is failing due to multiple warnings -> https://readthedocs.org/projects/nautobot-plugin-circuit-maintenance/builds/19862799/

cmsirbu:
Fixes #245 

Added missing files to mkdocs nav index.
Fixed a bunch of `mkdocstrings/griffe` warnings.

This is the only solution I could find - I think it doesn't like the fact that there's no explicit constructor to map to the `Args` piece. It's not exactly a big loss I think... but perhaps someone has a better solution?

```
-    Args:
-        Job (Nautobot Job): Nautobot Job parent class
```

Error being

```
Parameter 'Job' does not appear in the function signature
```

### DIFF
--- a/nautobot_circuit_maintenance/jobs/site_search.py
+++ b/nautobot_circuit_maintenance/jobs/site_search.py
@@ -15,7 +15,7 @@ CIRCUIT_MAINTENANCE_TAG_COLOR = "Purple"
 name = "Circuit Maintenance"  # pylint: disable=invalid-name
 
 
-def check_for_overlap(record1: CircuitMaintenance, record2: CircuitMaintenance):
+def check_for_overlap(record1: CircuitMaintenance, record2: CircuitMaintenance) -> bool:
     """Checks for the overlap of two circuit maintenance records.
 
     Args:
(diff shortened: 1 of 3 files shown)


nautobot_circuit_maintenance/jobs/site_search.py
```diff
@@ -94,9 +94,6 @@ class FindSitesWithMaintenanceOverlap(Job):
 
     Future iterations may include the ability to search for multiple circuit overlaps that would allow for just a single
     circuit to be available.
-
-    Args:
-        Job (Nautobot Job): Nautobot Job parent class
```
glennmatthews, line 99: FWIW I think this is correct to remove. The class inheritance isn't an `arg` to this class's constructor.

glennmatthews (approved):

PR #267, opened 2023-09-13
fix: :heavy_minus_sign: Remove google-auth dependency

chadell:


changes/267.fixed
```diff
@@ -0,0 +1 @@
+Fixed (by removing) dependency `google-auth`
```
snaselj, line 1:
```suggestion
Fixed (by removing) dependency `google-oauth`.
```

snaselj (approved):
Change fragment can be updated, otherwise LGTM

PR #272, opened 2023-09-28
Fix test in next-2.0

glennmatthews:
2.0 recently (https://github.com/nautobot/nautobot/pull/4443) added a test case to the generic API that expects three object instances to be present, but this test was only creating two instances. We should fix the test to fail more gracefully when that is the case, but the changes here should get tests back to passing for this app.

(edit) After fixing the number of test objects, this test case uncovered a legitimate issue in this app that several bulk-operation views were also missing filterset declarations.

### DIFF
--- a/nautobot_circuit_maintenance/views.py
+++ b/nautobot_circuit_maintenance/views.py
@@ -247,6 +247,7 @@ class CircuitMaintenanceBulkEditView(generic.BulkEditView):
     """View for bulk editing circuitmaintenance features."""
 
     queryset = models.CircuitMaintenance.objects.all()
+    filterset = filters.CircuitMaintenanceFilterSet
     table = tables.CircuitMaintenanceTable
     form = forms.CircuitMaintenanceBulkEditForm
 
@@ -255,6 +256,7 @@ class CircuitMaintenanceBulkDeleteView(generic.BulkDeleteView):
     """View for bulk deleting circuitmaintenance features."""
 
     queryset = models.CircuitMaintenance.objects.all()
+    filterset = filters.CircuitMaintenanceFilterSet
     table = tables.CircuitMaintenanceTable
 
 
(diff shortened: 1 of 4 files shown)


glennmatthews:
Naturally that wasn't the only failure. Working to address the remaining ones.

itdependsnetworks:
@glennmatthews this is more for core, but I have been thinking about this, since this has broken tests in the past, I think it makes sense to implement this simple pattern, and remove the defaulting every major semver change.


```python
    class CreateObjectViewTestCase(ModelViewTestCase):
        """
        Create a single new instance.

        :form_data: Data to be used when creating a new object.
        """

        do_create_object_without_permission = False


        def test_create_object_without_permission(self):
            # Try GET without permission
            if not self.do_create_object_without_permission:
                pass
            # perform actual test
```

Ideally coupled with pylint, it informs people to turn them on, without hard failing during minor and maybe even patch version updates. I do realize it is more work to "turn it back on" all of the models, but it goes to the "supported public interface" convo.

glennmatthews:
In general I'd disagree - most of the time these tests are uncovering existing (latent) bugs in the apps; it's rare that the failures are just pure noise.

glennmatthews:
Case in point here - the new tests uncovered that a number of Bulk operation views were missing a filterset declaration and so couldn't accept a filtered list of objects to bulk edit/delete.

itdependsnetworks:
apologies if it seemed like I implied they are noise, it is simply about the public interface. We can take a stance as you are seemingly suggesting that this is to be expected, but I wouldn't have expected my CI to break over minor semver upgrades.

nautobot_circuit_maintenance/views.py
```diff
@@ -372,6 +375,7 @@ class NoteBulkDeleteView(generic.BulkDeleteView):
     """View for bulk deleting Notea."""
```
HanlinMiao, line 375:
```suggestion
    """View for bulk deleting Notes."""
```

HanlinMiao (approved):

PR #290, opened 2024-03-15
Ltm 1.6 - Pydantic 1.x and 2.x support

chadell:


.github/workflows/ci.yml
```diff
@@ -121,7 +121,7 @@ jobs:
       matrix:
         python-version: ["3.8", "3.9", "3.10"]
         db-backend: ["postgresql"]
-        nautobot-version: ["stable"]
+        nautobot-version: ["1.6.14"]
```
glennmatthews, line 124: can update to 1.6.15 now, but that's not critically important.

pyproject.toml
```diff
@@ -26,6 +25,7 @@ markdown = "!=3.3.5"  # Added due an issue with Markdown 3.3.5
 nautobot = { version = "^1.4.0", optional = true }  # Used for local development
 nautobot-capacity-metrics = {version = "*", optional = true }
 python = "^3.8"
+pydantic = "^2.0.0"
```
glennmatthews, line 28: LTM requires pydantic 2.x at this point?

chadell: if we want to keep using the latest updates from circuit-maintenance-parser, yes. All the users of Nautobot 1.6, without this, won't be able to take the new parsers without this bump

glennmatthews (approved):

PR #292, opened 2024-03-20
Release v2.1.0

chadell:


LICENSE
```diff
@@ -1,7 +1,5 @@
 
-                                 Apache License
-                           Version 2.0, January 2004
-                        http://www.apache.org/licenses/
+Copyright (c) 2024, Network to Code, LLC
```
glennmatthews, line 2: Seems like a weird change to remove the "Apache License" from this file. Is this from drift manager?

glennmatthews (approved):

LICENSE
chadell: yes, I think that all this changes are coming from drift manager

PR #300, opened 2024-05-31
Update develop to 2.2.0

mattmiller87:


glennmatthews (commented):
Probably need to run `towncrier build --version 2.1.1` to get the release-notes updated.

poetry.lock
glennmatthews, line 1: What was the reason for regenerating this file? Shouldn't need to be touched as a part of the release prep.

jvanderaa:
That towncrier part was new. I thought it was still `poetry version patch`. Will check on this.

chadell:
We should do a minor version because we are adding new features not only patches

glennmatthews (approved):

PR #310, opened 2024-06-28
Update version with patch.

jvanderaa:
Sets patch release.


jvanderaa:
Fixes #311

glennmatthews (commented):
Although it's not spelled out yet in the project docs so far as I can see, I think the usual process for a release would be to do a release-branch PR into `main` (including release-note updates, etc.) rather than directly bumping the version in `develop`. Happy to collaborate with you on this if needed.

jvanderaa:
So I would cut a new branch off of develop, PR that to main, then merge main back to develop?

glennmatthews:
That's how we do things in core at least.

jvanderaa:
I can do that then. What I've been doing on pynautobot, nautobot-ansible is:

merge pieces into develop
when ready PR develop into main and then cut release.

chadell:
I guess the changelog is failing because the `release` file? is the first time I see it and I have not used it before for the releases

glennmatthews:
I see that the target of the PR was changed correctly to `main`; in that case we need to run `towncrier` to get the *existing* change fragments collected into the release-note.

jvanderaa:
Ok, this was a case of not following the new documentation. I have some questions on it with one change that we should make. The part about tags. It says to create a tag. Shouldn't that be `Create a Release`? What are the differences?

jvanderaa:
@chadell I think this should be all set.

changes/311.release
```diff
@@ -0,0 +1 @@
+Released 2.2.3.
```
glennmatthews, line 1: This file shouldn't be needed for a PR into `main`. The documentation isn't as clear as it could be.

glennmatthews (approved):
The release process after this is merged should indeed be to create a `Release` and in the process of doing so, have GitHub create a `v2.2.3` `tag` for the release.

changes/311.release
jvanderaa: It was left over. Documentation (which I will tackle, you know me now :) ) was fine.

chadell (approved):
LGTM

PR #321, opened 2024-09-12
Main to develop post 2.2.4

smk4664:
This pr merges the 2.2.4 release into develop.


chadell (approved):
LGTM

smk4664:
This is missing the `poetry version prerelease`, but that is because I am going to immediately open a release branch for Circuit Maintenance 2.3.0 which will be adding Python 3.12 support.

Issue #278, opened 2023-12-07
GmailAPI needs retry logic for intermittent HttpError


glennmatthews:
### Environment
* Python version:  
* Nautobot version:   1.6.6
* nautobot-circuit-maintenance version:   1.0.1

### Expected Behavior

Update Circuit Maintenance job should run to completion when retrieving emails from Gmail.

### Observed Behavior

The `GmailAPI.fetch_email()` function doesn't have any try/except logic for `googleapiclient.errors.HttpError` exceptions, so when one happens, the entire job fails and aborts.

This is not a "hard" failure like bad authentication credentials, this instead happens partway through an otherwise successful job run where any number of emails are initially retrieved and processed, but for one particular email an intermittent HttpError exception is encountered:

![image](https://github.com/nautobot/nautobot-plugin-circuit-maintenance/assets/5603551/163c2dfb-7879-4a10-9512-05f4de79ac83)

### Steps to Reproduce

1.
2.
3.

chadell:
fixed in #323

PR #323, opened 2024-09-13
fix: :bug: Provide retry logic for GoogleAPI execute

chadell:
Fixes issue #278

### DIFF
--- a/nautobot_circuit_maintenance/handle_notifications/sources.py
+++ b/nautobot_circuit_maintenance/handle_notifications/sources.py
@@ -8,6 +8,7 @@ import json
 import logging
 import os
 import re
+import time
 from typing import Dict, Iterable, List, Optional, Tuple, Type, TypeVar, Union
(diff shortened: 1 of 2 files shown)


gsnider2195:
This needs to be ruff formatted. Sorry about the problem with the check not failing like it's supposed to 😞

gsnider2195 (approved):
Other than the ruff comment this looks great!

changes/323.fixed
```diff
@@ -0,0 +1 @@
+Provide retry logic for GoogleAPI execute
```
gsnider2195, line 1:
```suggestion
Added retry logic to GoogleAPI execute.
```

chadell:
> This needs to be ruff formatted. Sorry about the problem with the check not failing like it's supposed to 😞

in the CI, the `ruff-format` before the `ruff-lint`, seems to hide the problem, isn't it?

PR #325, opened 2024-09-16
Merge to develop after release v2.3.0

chadell:


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "nautobot-circuit-maintenance"
-version = "2.2.4"
+version = "2.3.0"
```
glennmatthews, line 3: bump to `2.3.1b1` perhaps?

glennmatthews (approved):

pyproject.toml
chadell: do you mean to prepare for the next release? I haven't seen this in the process, but it could be a good idea to consider

PR #336, opened 2025-04-01
Do not leak secrets in logs

chadell:
# Closes: #<ISSUE NUMBER GOES HERE>

## What's Changed

- Removed the password and credentials from the serialization of the `Source` object.
- Simplified the log message for a notification to only relevant data.

## To Do

- [ ] Explanation of Change(s)
- [ ] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/#creating-changelog-fragments))
- [ ] Attached Screenshots, Payload Example
- [ ] Unit, Integration Tests
- [ ] Documentation Updates (when adding/changing features)
- [ ] Outline Remaining Work, Constraints from Design


changes/336.fixed
```diff
@@ -0,0 +1,2 @@
+Removed the password and credentials from the serialization of the Source object.
```
glennmatthews: Should this be a `.security` note?

nautobot_circuit_maintenance/handle_notifications/handler.py
```diff
@@ -433,7 +433,9 @@ def run(self, dryrun=False) -> List[uuid.UUID]:
 
         raw_notification_ids = []
         for notification in notifications:
-            self.logger.info(f"Processing notification `{notification.subject}`.", extra={"object": notification})
+            self.logger.info(
+                f"Processing notification `{notification.subject}` with ID `{notification.msg_id}` from source `{notification.source.name}`."
+            )
```
glennmatthews, line 438: This change removes the link to the `notification` object so that it's no longer possible to click through from the log message to the notification itself. Is that desirable?

nautobot_circuit_maintenance/handle_notifications/sources.py
```diff
@@ -534,7 +534,7 @@ class GmailAPI(EmailSource):
     credentials_file: str
     account: str
```
glennmatthews, line 535: For my education, why doesn't `account` need a `repr=False` declaration?

changes/336.fixed
chadell: good point, I wasn't aware of it

nautobot_circuit_maintenance/handle_notifications/handler.py
chadell: the Notification object is not a Nautobto object, so it's not linked and only a partial representation is shown.

nautobot_circuit_maintenance/handle_notifications/sources.py
chadell: I considered the account as not so sensitive, I could include, for sure

glennmatthews: I was going by the fact that the tests are asserting the account is also not present in the repr.

chadell: well, I'm testing the opposite
image

glennmatthews: Sorry, reading failure on my part :)

glennmatthews (approved):

PR #344, opened 2025-09-10
Refactored CircuitMaintenanceUIViewSet, CircuitImpactUIViewSet, NoteUIViewSet, N…

Pavan-183:
| View                                        | Before                                                                                                                                                                                   | After                                                                                                                                                                                    |
| ------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| **List View (Circuit Maintenances)**        | image <br> image | image <br> image |
| **Detail View (Circuit Maintenances)**      | image                                                                                                | image <br>image                                                                                                
| **Update View (Circuit Maintenances)**      | image <br> image | image <br> image |
| **Delete View (Circuit Maintenances)**      | image <br> image | image <br> image |
| **Bulk Edit View (Circuit Maintenances)**   | image <br> image | image <br> image |
| **Bulk Delete View (Circuit Maintenances)** | image <br> image | image <br> image |
| **Add View (Circuit Maintenances)**         | image <br> image | image <br> image |
| **Add View (Circuit Impact)**               | image <br> image | image <br> image |
| **Add View (Note)**                         | image <br> image | image <br> image |
| **List View (Notifications)**               | image <br> image | image <br> image |
| **Detail View (Notifications)**             | image                                                                                                | image                                                                                               |
| **List View (Notification Source)**         | image                                                                                                | image                                                                                                |
| **Detail View (Notification Source)**       | image                                                                                                | image                                                                                                |
| **Update View (Notification Source)**       | NA                                                                                                                                                                                       | image <br> image |
| **Bulk Update View (Notification Source)**  | image                                                                                                | image                                                                                                |
| **Bulk Delete View (Notification Source)**  | NA                                                                                                                                                                                       | image <br> image |
| **Delete View (Notification Source)**       | NA                                                                                                                                                                                       | image |

# Closes: #<ISSUE NUMBER GOES HERE>

## What's Changed

- Upgraded Nautobot dependency to v2.4.16 across code, configs, and docs.
- Refactored CircuitMaintenance, CircuitImpact, Note, NotificationSource model related UI views to use `NautobotUIViewSet`.
- To make the test cases pass, I added actions for `validate_source` however, I’m not sure if this is the correct approach and will clarify with the apps team.

## To Do

- [x] Explanation of Change(s)
- [x] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/core/#creating-changelog-fragments))
- [x] Attached Screenshots, Payload Example
- [ ] Unit, Integration Tests
- [ ] Documentation Updates (when adding/changing features)
- [ ] Outline Remaining Work, Constraints from Design


Pavan-183:
image
I have added the banner, but the test cases are failing, so I have left it as it is for now.(will confirm this with apps team.)

development/nautobot_config.py
```diff
@@ -10,6 +10,14 @@
...
 # Debug
 #
+# Ensure debug_toolbar app is registered if the package is installed on disk
+try:
+    import debug_toolbar  # noqa: F401
+except ImportError:
+    pass
+else:
+    if "debug_toolbar" not in INSTALLED_APPS:  # noqa: F405
+        INSTALLED_APPS.append("debug_toolbar")  # noqa: F405
```
Sridevi-NTC: Please remove these lines of code and downgrade debug_toolbar version

Pavan-183: sure @Sridevi-NTC

Pavan-183:
Updated as requested, should be resolved, let me know if I missed anything

jvanderaa:
What is the purpose of the refactor?

Pavan-183:
@jvanderaa , The purpose of the refactor is to consolidate the multiple individual views (edit, delete, bulk actions, etc.) into a single `UIViewSet`. This reduces duplicated code, makes the views easier to maintain, and leverages Nautobot's standardized `UIViewSet` pattern for consistency.

nautobot_circuit_maintenance/forms.py
```diff
@@ -55,7 +51,7 @@ class Meta:  # noqa: D106 "Missing docstring in public nested class"
         nullable_fields = ["impact"]
 
 
-class CircuitImpactFilterForm(BootstrapMixin, CustomFieldModelFilterFormMixin):
+class CircuitImpactFilterForm(NautobotFilterForm):
```
jvanderaa, line 56: I'm not really understanding why this change. I search for `NautobotModelForm` on the Nautobot Docs site and nothing comes up.

nautobot_circuit_maintenance/forms.py
```diff
@@ -73,18 +69,18 @@ class CircuitImpactFilterForm(BootstrapMixin, CustomFieldModelFilterFormMixin):
...
-class CircuitMaintenanceForm(BootstrapMixin, CustomFieldModelFormMixin, RelationshipModelFormMixin):
+class CircuitMaintenanceForm(NautobotModelForm):
     """Filter Form for CircuitMaintenance instances."""
 
     class Meta:  # noqa: D106 "Missing docstring in public nested class"
         """Metaclass attributes for CircuitMaintenanceAddForm."""
 
         model = CircuitMaintenance
-        fields = ["name", "start_time", "end_time", "description", "status", "ack"]
+        fields = "__all__"
```
jvanderaa, line 81: Why this change?

nautobot_circuit_maintenance/forms.py
```diff
@@ -107,30 +103,32 @@ class CircuitMaintenanceFilterForm(BootstrapMixin, CustomFieldModelFilterFormMix
     end_time = forms.DateTimeField(label="End time before", required=False, widget=DateTimePicker())
 
 
-class CircuitMaintenanceBulkEditForm(BootstrapMixin, AddRemoveTagsForm, CustomFieldModelBulkEditFormMixin):
+class CircuitMaintenanceBulkEditForm(NautobotBulkEditForm, AddRemoveTagsForm):
```
jvanderaa: More what is this about?

nautobot_circuit_maintenance/templates/nautobot_circuit_maintenance/circuitimpact.html
```diff
@@ -1,27 +1,2 @@
...
-                </td>
-            </tr>
-            <tr>
-                <td>Impact</td>
-                <td>{{ object.impact }}</td>
-            </tr>
-        </table>
-    </div>
-{% endblock %}
+{% extends 'nautobot_circuit_maintenance/circuitimpact_retrieve.html' %}
```
jvanderaa, line 1: Why this change?

nautobot_circuit_maintenance/templates/nautobot_circuit_maintenance/circuitmaintenance.html
```diff
@@ -1,221 +1,2 @@
...
-                        <pre>{{ parsed_notification.json | render_json }}</pre>
-                    </td>
-                </tr>
-                {% endfor %}
-            </tbody>
-        </table>
-        {% endif %}
-    </div>
-{% endblock %}
+{% extends 'nautobot_circuit_maintenance/circuitmaintenance_retrieve.html' %}
```
jvanderaa, line 1: Not understanding this change.

nautobot_circuit_maintenance/views.py
```diff
@@ -273,156 +239,76 @@ def get(self, request, *args, **kwargs):
...
 
-class RawNotificationDeleteView(generic.ObjectDeleteView):
-    """View for deleting Raw Notification."""
+    def _process_bulk_update_form(self, form):
+        pass
 
-    model = models.RawNotification
-    queryset = models.RawNotification.objects.all()
+    def _process_create_or_update_form(self, form):
+        pass
```
jvanderaa: Not understanding why the numerous creation of things with just a pass on it.

pyproject.toml
```diff
@@ -44,7 +44,7 @@ pydantic = ">=1.10.4,<3"
 
 [tool.poetry.group.dev.dependencies]
 coverage = "*"
-django-debug-toolbar = "*"
+django-debug-toolbar = "5.2.0"
```
jvanderaa: Why this change? I think this will come back to cause more work later.

jvanderaa (commented):
There is a lot going on in this one change. Should probably be smaller changes going forward.

nautobot_circuit_maintenance/forms.py
Pavan-183: This change replaces the direct use of multiple mixins with `NautobotFilterForm`, which is a convenience base class provided in the codebase. It already combines BootstrapMixin, CustomFieldModelFilterFormMixin, and RelationshipModelFilterFormMixin (and in some cases ContactTeamModelFilterFormMixin).


```
class NautobotFilterForm(
    BootstrapMixin,
    # The below must be listed *after* BootstrapMixin so that BootstrapMixin applies to their dynamic form fields
    ContactTeamModelFilterFormMixin,
    CustomFieldModelFilterFormMixin,
    RelationshipModelFilterFormMixin,
):
    """
    This class exists to combine common functionality and is used to inherit from throughout the
    codebase where all of ContactTeamModelFilterFormMixin, CustomFieldModelFilterFormMixin, and
    RelationshipModelFilterFormMixin are needed.
    """
```

nautobot_circuit_maintenance/forms.py
Pavan-183: I made this change because pylint flagged it, it recommended using `__all__ ` instead of explicitly listing every field.

nautobot_circuit_maintenance/forms.py
Pavan-183: This change switches from manually combining multiple mixins to using `NautobotBulkEditForm`, which is the built-in base form that already includes `BootstrapMixin`, `CustomFieldModelBulkEditFormMixin`, `RelationshipModelBulkEditFormMixin`, and `NoteModelBulkEditFormMixin`. It reduces duplication and makes the form consistent with other bulk edit forms across Nautobot.

```
class NautobotBulkEditForm(
    BootstrapMixin,
    # The below must be listed *after* BootstrapMixin so that BootstrapMixin applies to their dynamic form fields
    CustomFieldModelBulkEditFormMixin,
    NoteModelBulkEditFormMixin,
    RelationshipModelBulkEditFormMixin,
):
    """Base class for bulk-edit forms for models that support relationships, custom fields and notes."""

```

nautobot_circuit_maintenance/views.py
Pavan-183: These stub methods were added because `pylint` was flagging missing required method definitions for the base class. Since we don’t have any custom logic to implement for these methods right now, they just contain pass. They’re essentially placeholders to satisfy the linter and keep the class structure aligned with Nautobot’s expected interface.

pyproject.toml
Pavan-183: We were hitting following error

`RuntimeError: Model class debug_toolbar.models.HistoryEntry doesn't declare an explicit app_label and isn't in an application in INSTALLED_APPS. `

with the unpinned version. Pinning to 5.2.0 resolved the issue and allowed things to run correctly.

nautobot_circuit_maintenance/templates/nautobot_circuit_maintenance/circuitimpact.html
Pavan-183: In core we’ve been [following this pattern](https://github.com/nautobot/nautobot/pull/7160/files#diff-921d054d4d5ba5707035451fd5e86df8f17d2160b643b89a00b370c910169bc6) for templates, so I updated this to stay consistent with that approach.

nautobot_circuit_maintenance/templates/nautobot_circuit_maintenance/circuitmaintenance.html
Pavan-183: In core we’ve been [following this pattern](https://github.com/nautobot/nautobot/pull/7160/files#diff-921d054d4d5ba5707035451fd5e86df8f17d2160b643b89a00b370c910169bc6) for templates, so I updated this to stay consistent with that approach.

nautobot_circuit_maintenance/forms.py
jvanderaa: Do you have a link to the documentation on this FIlterForm?

nautobot_circuit_maintenance/forms.py
jvanderaa: I would have expected pylint to have flagged this before. This PR isn't updating pylint.

nautobot_circuit_maintenance/templates/nautobot_circuit_maintenance/circuitimpact.html
jvanderaa: I'll need some input from @glennmatthews or @gsnider2195 on this one. I don't understand the need for the change.

nautobot_circuit_maintenance/views.py
jvanderaa: Why is pylint suddenly catching this?

pyproject.toml
jvanderaa: This should install from the poetry.lock file, leaving it at the last known version. So we shouldn't be needing to pin it. Unless I'm missing something in the past 6 months here. @glennmatthews can you validate?

gsnider2195: DJDT version 6 introduced a model in that library which broke it in all of our apps. We've been pinning it to `<6.0.0` until we raise the minimum of Nautobot to a version that either fixes the issue or pins the version for us.

nautobot_circuit_maintenance/templates/nautobot_circuit_maintenance/circuitimpact.html
glennmatthews: It's a pain, frankly. :) But basically the old generic views look for `<model>.html` by default but UIViewSet looks for `<model>_retrieve.html` by default. I'd proposed adding a fallback to UIViewSet to make these changes unnecessary but it never went anywhere so here we are.

nautobot_circuit_maintenance/views.py
glennmatthews: This seems suspicious to me. Are we using the wrong set of mixins for the viewset by chance?

nautobot_circuit_maintenance/forms.py
Pavan-183: Yes @jvanderaa ,it’s [documented here ](https://docs.nautobot.com/projects/core/en/stable/code-reference/nautobot/apps/forms/#nautobot.apps.forms.NautobotFilterForm)

nautobot_circuit_maintenance/views.py
Pavan-183: These mixins (ObjectBulkDestroyViewMixin, ObjectDestroyViewMixin, etc.) are coming from nautobot.core.views.mixins
@glennmatthews 

```
from nautobot.core.views.mixins import (
    ObjectBulkDestroyViewMixin,
    ObjectDestroyViewMixin,
    ObjectDetailViewMixin,
    ObjectListViewMixin,
)
```

glennmatthews: Are they the right mixins for this specific view though?

nautobot_circuit_maintenance/forms.py
jvanderaa: It looks like the NautobotFilterForm has 4 bases and the previous one had 2. Any concerns here?

pyproject.toml
jvanderaa: I'm good with `<6.0.0`. That makes sense. `="5.2.0"` does not.

nautobot_circuit_maintenance/templates/nautobot_circuit_maintenance/circuitimpact.html
jvanderaa: Thanks @glennmatthews that explanation makes sense.

nautobot_circuit_maintenance/views.py
Pavan-183: yes @glennmatthews

glennmatthews: I'm not understanding why a view with those mixins should need for example an implementation of `_process_bulk_update_form` when we're not using `ObjectBulkUpdateMixin` and so that method will never be called.

Pavan-183: When I comment out those methods and run pylint and the tests, I get errors. I'm not sure why, @glennmatthews . Since these abstract methods aren’t needed here, is it okay if I disable pylint for this issue, or would you prefer I keep stub implementations?

image

glennmatthews: Better to disable the pylint warning here (`pylint: disable=abstract-method`) then add stub code that provides no value.

Pavan-183: Sure @glennmatthews

nautobot_circuit_maintenance/forms.py
itdependsnetworks: No, it's a net positive. We now do relationship and contact team filters.

nautobot_circuit_maintenance/forms.py
itdependsnetworks: Pavan's local instance may have bumped it. @Pavan-183 can you just confirm there is nothing odd about this, such as "start_time" is a method or something.

nautobot_circuit_maintenance/urls.py
```diff
@@ -3,71 +3,25 @@
...
 app_name = "nautobot_circuit_maintenance"
-# TODO: Uncomment and add to urlpatterns when converting to UIViewSets.
-# router = NautobotUIViewSetRouter()
+
+router = NautobotUIViewSetRouter()
+router.register("impact", views.CircuitImpactUIViewSet)
+router.register("maintenance", views.CircuitMaintenanceUIViewSet)
+router.register("note", views.NoteUIViewSet)
+router.register("rawnotification", views.RawNotificationUIViewSet)
+router.register("source", views.NotificationSourceUIViewSet)
```
itdependsnetworks, line 17:
```suggestion
router.register("circuit-impact", views.CircuitImpactUIViewSet)
router.register("circuit-maintenance", views.CircuitMaintenanceUIViewSet)
router.register("note", views.NoteUIViewSet)
router.register("raw-notification", views.RawNotificationUIViewSet)
router.register("notification-source", views.NotificationSourceUIViewSet)
```

glennmatthews:
```suggestion
router.register("circuit-impacts", views.CircuitImpactUIViewSet)
router.register("circuit-maintenances", views.CircuitMaintenanceUIViewSet)
router.register("notes", views.NoteUIViewSet)
router.register("raw-notifications", views.RawNotificationUIViewSet)
router.register("notification-sources", views.NotificationSourceUIViewSet)
```

nautobot_circuit_maintenance/views.py
itdependsnetworks: As long as it works... I was concerned this was being called anyway. If we can remove, please inform the rest of the team to remove across the board.

nautobot_circuit_maintenance/urls.py
gsnider2195: Aren't these breaking changes? If we change the urls here shouldn't they match `api/urls.py`?

itdependsnetworks: Making me eat my own words.. huh :) 

image


I do think it makes sense to do a redirect, but I think fair enough to consider outside of scope.

nautobot_circuit_maintenance/views.py
```diff
@@ -431,83 +319,83 @@ class ParsedNotificationView(generic.ObjectView):
...
+            except ValueError as error:
+                context.update(
+                    {
+                        "providers": Provider.objects.filter(
+                            pk__in=[provider.pk for provider in instance.providers.all()]
+                        ),
+                        "account": str(error),
+                        "source_type": str(error),
+                    }
+                )
```
itdependsnetworks: What is this for?

This also seems like left over from troubleshooting: `"account": str(error),`

nautobot_circuit_maintenance/views.py
```diff
@@ -431,83 +319,83 @@ class ParsedNotificationView(generic.ObjectView):
...
+
+    @action(
+        detail=True,
+        methods=["get"],
+        url_path="validate",
+        url_name="validate",
+        custom_view_base_action="view",
+        custom_view_additional_permissions=["nautobot_circuit_maintenance.view_notificationsource"],
+    )
+    def validate_source(self, request, pk=None):  # pylint: disable=unused-argument
```
itdependsnetworks, line 366: Does this view need this context?

```
        return {
            "authentication_message": message,
            "providers": Provider.objects.filter(pk__in=[provider.pk for provider in instance.providers.all()]),
            "account": source.get_account_id(),
            "source_type": source.__class__.__name__,
            "active_tab": "main",
        }
```

nautobot_circuit_maintenance/forms.py
Pavan-183: @itdependsnetworks , Confirmed, `start_time` (and the other listed attributes) are standard model fields on `CircuitMaintenance`, not methods or properties.

nautobot_circuit_maintenance/views.py
Pavan-183: yes @itdependsnetworks , previously due to this few test cases are failed, so I have removed it, now I have resolved this issue, will commit the updated code

nautobot_circuit_maintenance/views.py
Pavan-183: yes @itdependsnetworks , I have updated it now.

nautobot_circuit_maintenance/views.py
Pavan-183: Updated as requested, should be resolved, let me know if I missed anything

Pavan-183:
Updated as requested, should be resolved, let me know if I missed anything @itdependsnetworks

nautobot_circuit_maintenance/views.py
```diff
@@ -5,16 +5,28 @@
...
+    ObjectListViewMixin,
+)
+from rest_framework.decorators import action
+from rest_framework.response import Response
 
 from nautobot_circuit_maintenance import filters, forms, models, tables
 from nautobot_circuit_maintenance.handle_notifications.sources import RedirectAuthorize, Source
 from nautobot_circuit_maintenance.models import CircuitMaintenance
 
+from .api import serializers
```
itdependsnetworks: Use the full reference. Usually we do

```
from nautobot_circuit_maintenance import filters, forms, models, tables
from nautobot_circuit_maintenance.api import serializers
```

can remove this as well: 
`from nautobot_circuit_maintenance.models import CircuitMaintenance` and replace with `models.CircuitMaintenance`

Pavan-183: sure @itdependsnetworks

Pavan-183: Updated as requested, should be resolved, let me know if I missed anything

nautobot_circuit_maintenance/views.py
Pavan-183: Updated as requested, should be resolved, let me know if I missed anything

nautobot_circuit_maintenance/views.py
```diff
@@ -5,15 +5,25 @@
...
-from nautobot.circuits.models import Circuit, Provider
+from nautobot.apps.views import NautobotUIViewSet
+from nautobot.circuits.models import Circuit
 from nautobot.core.views import generic
+from nautobot.core.views.mixins import (
+    ObjectBulkDestroyViewMixin,
+    ObjectDestroyViewMixin,
+    ObjectDetailViewMixin,
+    ObjectListViewMixin,
+)
```
itdependsnetworks: Can import all of these and the generic to import from `nautobot.app.views`.  Circuits you can't, just as a heads up.

Pavan-183: Updated as requested, should be resolved, let me know if I missed anything

nautobot_circuit_maintenance/views.py
```diff
@@ -5,15 +5,25 @@
...
+from nautobot.apps.views import (
+    NautobotUIViewSet,
+    ObjectBulkDestroyViewMixin,
+    ObjectDestroyViewMixin,
+    ObjectDetailViewMixin,
+    ObjectListViewMixin,
+)
+from nautobot.circuits.models import Circuit
 from nautobot.core.views import generic
+from rest_framework.decorators import action
```
itdependsnetworks, line 22:
```suggestion
from nautobot.apps.views import (
    NautobotUIViewSet,
    ObjectBulkDestroyViewMixin,
    ObjectDestroyViewMixin,
    ObjectDetailViewMixin,
    ObjectListView,
    ObjectListViewMixin,
    ObjectView
)
from nautobot.circuits.models import Circuit
from rest_framework.decorators import action
```

itdependsnetworks: need to removed the `generic.ObjectView` and `generic.ObjectView` to ones without the word generic in them.

Pavan-183: Okay @itdependsnetworks

Pavan-183: Updated as requested, should be resolved, let me know if I missed anything

nautobot_circuit_maintenance/forms.py
```diff
@@ -107,30 +103,32 @@ class CircuitMaintenanceFilterForm(BootstrapMixin, CustomFieldModelFilterFormMix
...
     """Form for bulk editing Circuit Maintenances."""
 
     pk = forms.ModelMultipleChoiceField(queryset=CircuitMaintenance.objects.all(), widget=forms.MultipleHiddenInput)
-    status = forms.CharField(max_length=200, required=False)
+    status = forms.ChoiceField(choices=CircuitMaintenanceStatusChoices, required=False)
```
jvanderaa: Does this change possibly break something that folks have relied on. If so, we should make sure that we have the appropriate change documentation. Some may be using the CharField that we had previously.

nautobot_circuit_maintenance/forms.py
```diff
@@ -107,30 +103,32 @@ class CircuitMaintenanceFilterForm(BootstrapMixin, CustomFieldModelFilterFormMix
...
     ack = forms.BooleanField(required=False, widget=BooleanWidget())
     description = forms.CharField(max_length=200, required=False)
 
     class Meta:  # noqa: D106 "Missing docstring in public nested class"
-        nullable_fields = ["status", "ack", "description"]
+        # Only description is nullable in the model.
+        # status and ack were removed from nullable_fields because tests were failing.
+        nullable_fields = ["description"]
```
jvanderaa, line 119: `ack` should be allowed to be nullable.

jvanderaa:
@itdependsnetworks I'm going to ask for a little bit of your guidance on these last two comments that I made here. Thoughts?

nautobot_circuit_maintenance/forms.py
gsnider2195: `ack` was made non-nullable in https://github.com/nautobot/nautobot-app-circuit-maintenance/pull/261
I'm not sure if that was intentional though.

nautobot_circuit_maintenance/forms.py
itdependsnetworks: I don't think this can be considered breaking... if somehow they bypassed the model, I suspect other places in the code are enforcing what the models states:

https://github.com/nautobot/nautobot-app-circuit-maintenance/blob/4043fe1d70b20f13510160d62348782aef1a2156/nautobot_circuit_maintenance/models.py#L48-L52

itdependsnetworks:
> @itdependsnetworks I'm going to ask for a little bit of your guidance on these last two comments that I made here. Thoughts?

@jvanderaa I think @gsnider2195 answered the one and I got the other.

nautobot_circuit_maintenance/forms.py
jvanderaa: Interesting that drfit manager would change that.

jvanderaa (approved):

jvanderaa:
@Pavan-183 let me know if you are ready for this to be merged.

Pavan-183:
@jvanderaa , Just wanted to confirm, if everything looks good on your end and no one has any questions, we can proceed.

nautobot_circuit_maintenance/views.py
gsnider2195: It looks like this change may have gotten lost in a force push. I'm not seeing this in the branch and don't see any commits on this file after Oct 1.

development/nautobot_config.py
```diff
@@ -10,6 +10,7 @@
 #
 # Debug
 #
+# Ensure debug_toolbar app is registered if the package is installed on disk
```
gsnider2195:
```suggestion
```

development/nautobot_config.py
```diff
@@ -143,7 +144,7 @@
                 "name": "my imap source",
                 "account": os.environ.get("CM_NS_1_ACCOUNT", ""),
                 "secret": os.environ.get("CM_NS_1_SECRET", ""),
-                "url": os.environ.get("CM_NS_1_URL", ""),
+                "url": os.environ.get("CM_NS_1_URL", "https"),
```
gsnider2195: Is this in scope for the UIViewset changes?

development/nautobot_config.py
```diff
@@ -168,6 +169,13 @@
                 "credentials_file": os.environ.get("CM_NS_4_CREDENTIALS_FILE", ""),
                 # "attach_all_providers": True,
             },
+            {
```
gsnider2195: Is this in scope for the UIViewset changes?

nautobot_circuit_maintenance/urls.py
```diff
@@ -3,71 +3,25 @@
...
-    path("note/<uuid:pk>/changelog/", ObjectChangeLogView.as_view(), name="note_changelog", kwargs={"model": Note}),
-    # Raw Notification
-    path("rawnotification/", views.RawNotificationListView.as_view(), name="rawnotification_list"),
-    path("rawnotification/<uuid:pk>/", views.RawNotificationView.as_view(), name="rawnotification"),
-    path("rawnotification/<uuid:pk>/delete/", views.RawNotificationDeleteView.as_view(), name="rawnotification_delete"),
     path(
-        "rawnotification/delete/",
-        views.RawNotificationBulkDeleteView.as_view(),
-        name="rawnotification_bulk_delete",
+        "maintenance/job/", views.CircuitMaintenanceUIViewSet.as_view({"get": "run_job"}), name="circuitmaintenance_job"
```
gsnider2195: Do we need this? It looks like you're setting this url correctly in the `run_job` custom action.

nautobot_circuit_maintenance/urls.py
```diff
@@ -3,71 +3,25 @@
 from django.templatetags.static import static
 from django.urls import path
 from django.views.generic import RedirectView
-from nautobot.extras.views import ObjectChangeLogView
+from nautobot.core.views.routers import NautobotUIViewSetRouter
```
gsnider2195:
```suggestion
from nautobot.apps.urls import NautobotUIViewSetRouter
```

pyproject.toml
```diff
@@ -29,7 +29,7 @@ include = [
 [tool.poetry.dependencies]
 python = ">=3.9.2,<3.13"
 # Used for local development
-nautobot = ">=2.4.2,<3.0.0"
+nautobot = "^2.4.16"
```
gsnider2195: Trying to keep these compliant with PEP508

```suggestion
nautobot = ">=2.4.16,<3.0.0"
```

nautobot_circuit_maintenance/views.py
```diff
@@ -189,22 +199,19 @@ def get_maintenances_per_month(self):
...
-
-class CircuitMaintenanceView(generic.ObjectView):
-    """Detail view for specific circuit maintenances."""
-
-    queryset = models.CircuitMaintenance.objects.all()
-
-    def get_extra_context(self, request, instance):
+    def get_extra_context(self, request, instance=None):
         """Extend content of detailed view for Circuit Maintenance."""
         maintenance_note = models.Note.objects.filter(maintenance=instance)
```
gsnider2195, line 220: Are these making unnecessary database calls on non-detail views? Maybe we should put these in a `if instance is not None:` condition?

gsnider2195: I think a cleaner option may be
```py
if instance is None:
    return {}
```

development/nautobot_config.py
Pavan-183: No @gsnider2195 , I will revoke

development/nautobot_config.py
Pavan-183: No @gsnider2195 , I will revoke

nautobot_circuit_maintenance/urls.py
Pavan-183: Yes @gsnider2195 , I’ll remove it. Earlier, I added it because I ran into an issue, but I’ll remove it now.

nautobot_circuit_maintenance/views.py
Pavan-183: okay @gsnider2195

nautobot_circuit_maintenance/views.py
```diff
@@ -218,51 +229,9 @@ def get_extra_context(self, request, instance):
...
-
-
-class CircuitMaintenanceJobView(generic.ObjectView):
-    """Special View to trigger the Job to look for new Circuit Maintenances."""
-
-    queryset = models.CircuitMaintenance.objects.all()
-
-    def get(self, request, *args, **kwargs):
-        """Custom GET to run a the Job."""
+    @action(detail=False, methods=["get"], url_path="job", url_name="circuitmaintenance_job")
```
gsnider2195: The `circuitmaintenance_` part of the url gets added automatically so we need to change this to `url_name="job"`

```suggestion
    @action(detail=False, methods=["get"], url_path="job", url_name="job")
```

gsnider2195 (approved):

Issue #373, opened 2026-02-08
Bumps circuit-maintenance-parser from v2.9.0 to v2.10.0


michalis1:
### Environment
* Python version:  3.11
* Nautobot version:  3.0
* nautobot-circuit-maintenance version:  3.0.0

### Observed Behavior

Bump circuit-maintenance-parser to upgrade timezonefinder from its current outdated version to v8.2.0.

PR #375, opened 2026-02-08
Bump circuit-maintenance-parser from 2.9.0 to 2.12.0 (ltm)

michalis1:
# Closes: #373

## What's Changed

Bumps `circuit-maintenance-parser` from 2.9.0 to **2.12.0** in `poetry.lock`.

### Note on the target version

Issue #373 and the original version of this PR targeted 2.10.0. Two corrections were needed:

1. **The lock was not actually updated.** `poetry.lock` still resolved `circuit-maintenance-parser` to 2.9.0 — the same as the base `ltm-2.4` branch — despite the title and changelog fragment saying otherwise. The only lockfile changes were incidental regeneration noise (a `dill` marker tweak, some `PyYAML` cp38 wheel entries). This is consistent with `poetry lock` having been run instead of `poetry update circuit-maintenance-parser`, so Poetry kept the already-in-constraint 2.9.0 pin.

2. **2.10.0 was already several releases behind.** Since the update had to be redone anyway, it targets the latest release, 2.12.0.

### Lockfile changes

`pyproject.toml` is unchanged — the constraint is `^2.0.0`, so this is a lock-only update. Five packages moved, all within the parser's own dependency subtree:

| Package | From | To |
|---|---|---|
| `circuit-maintenance-parser` | 2.9.0 | 2.12.0 |
| `backoff` | 1.11.1 | 2.2.1 |
| `lxml` | 5.4.0 | 6.1.1 |
| `requests` | 2.32.5 | 2.34.2 |
| `urllib3` | 2.5.0 | 2.7.0 |

The `backoff` major bump is required by the parser itself — 2.12.0 declares `backoff = ">=2.2.1,<3.0.0"`. The rest refreshed within their existing constraints.

No packages were added or removed (216 before and after), and `poetry check --lock` passes.


jvanderaa:
Heads up — this PR doesn't actually appear to update `circuit-maintenance-parser`. The `poetry.lock` still resolves it to **2.9.0** (same as the base `ltm-2.4` branch), even though the title and the `changes/373.housekeeping` fragment say "2.9.0 → 2.10.0." The only real lockfile changes here are incidental regeneration noise (a `dill` marker tweak and some `PyYAML 6.0.3` cp38 wheel entries) — nothing touches the parser. Likely `poetry lock` was run instead of `poetry update circuit-maintenance-parser`, so Poetry kept the already-in-constraint 2.9.0 pin.

Also worth noting: the latest release is now **2.12.0** (2.10.0 is already a few versions behind). Since we'd be re-running the update anyway, do we want to bump straight to **2.12.0** instead of 2.10.0? Happy to redo the lock update either way — just let me know the target version.

PR #390, opened 2026-07-20
Rename Circuit Maintenance navigation menu group and list item

### LINKED ISSUE #389: Rename navigation menu group from "Circuit Maintenance App" to "Circuit Maintenances"
### Environment
* Nautobot version: N/A (UI label change)
* nautobot-circuit-maintenance version: current `develop`

### Proposed Functionality
The navigation menu group under the **Circuits** tab currently displays as **"Circuit Maintenance App"** (defined by the `NavMenuGroup` `name` in `nautobot_circuit_maintenance/navigation.py`). It should instead read **"Circuit Maintenances"** to describe the content of the menu rather than the app itself, consistent with how other Nautobot nav groups are named after the objects they contain.

**Change required:** `nautobot_circuit_maintenance/navigation.py:18` — `name="Circuit Maintenance App"` → `name="Circuit Maintenances"`.

**Consideration:** The group already contains a `NavMenuItem` named "Circuit Maintenances" (the list view, line 29). Renaming the group to the same string produces a group header and child item with identical text. Maintainers may prefer an alternative (e.g., keeping the group name distinct, or renaming the list item) — happy to adjust based on preference.

### Use Case
Clearer, object-oriented menu labeling that matches Nautobot navigation conventions, where menu groups are named for the records they expose rather than for the plugin providing them.

jvanderaa:
# Closes: #389

## What's Changed

Renames the app's navigation entries under the **Circuits** tab so they describe the
objects they expose rather than the plugin itself, matching Nautobot's navigation
conventions:

- **Menu group:** `Circuit Maintenance App` → `Circuit Maintenances`
- **List menu item:** `Circuit Maintenances` → `Maintenances`

Renaming the list item alongside the group also removes the redundancy that would
otherwise result in a group header and child item reading identically.

No functional changes — links, weights, and permissions are unchanged.

### Screenshots

| Before | After |
|--------|-------|
| _drop before image here_ | _drop after image here_ |

### Tests

Adds `tests/test_navigation.py`, which characterizes the full navigation structure
(tab, group, and every item's link/name/weight/permissions) and guards against
regression of both old names.

## To Do

- [x] Explanation of Change(s)
- [x] Added change log fragment(s)
- [x] Attached Screenshots, Payload Example
- [x] Unit, Integration Tests
- [x] Documentation Updates (when adding/changing features)
- [x] Outline Remaining Work, Constraints from Design

### Before

circuit_maintenance_nav_before

### After
circuit_maintenance_nav_after

### DIFF
--- a/nautobot_circuit_maintenance/navigation.py
+++ b/nautobot_circuit_maintenance/navigation.py
@@ -15,7 +15,7 @@ menu_items = (
         weight=NavigationWeightChoices.CIRCUITS,
         groups=(
             NavMenuGroup(
-                name="Circuit Maintenance App",
+                name="Circuit Maintenances",
                 weight=250,
                 items=(
                     NavMenuItem(
@@ -26,7 +26,7 @@ menu_items = (
                     ),
                     NavMenuItem(
                         link="plugins:nautobot_circuit_maintenance:circuitmaintenance_list",
-                        name="Circuit Maintenances",
+                        name="Maintenances",
                         weight=200,
                         permissions=["nautobot_circuit_maintenance.view_circuitmaintenance"],
                     ),
--- a/changes/389.changed
+++ b/changes/389.changed
@@ -1,0 +1,1 @@
+Renamed the navigation menu group from "Circuit Maintenance App" to "Circuit Maintenances" and the maintenance list menu item to "Maintenances".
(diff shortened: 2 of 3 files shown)


nautobot_circuit_maintenance/navigation.py
```diff
@@ -26,7 +26,7 @@
                     ),
                     NavMenuItem(
                         link="plugins:nautobot_circuit_maintenance:circuitmaintenance_list",
-                        name="Circuit Maintenances",
+                        name="Maintenances",
```
glennmatthews, line 29: Personally I'm on the fence about this change - it feels unnecessary to me.

PR #392, opened 2026-07-20
Add configurable lookback window to the Handle Notifications job (#170)

### LINKED ISSUE #170: Configurable lookback window
### Environment

- Nautobot version: 1.1.5

### Proposed Functionality

When running the handle_notifications job, allow the user to optionally specify a lookback window relative to the last previously processed notification, instead of hard-coding this window to 1 day.

### Use Case

For cases where an email or group of emails failed, and the parser libraries have now been updated to handle those emails, it would be useful to be able to manually re-run the job with a lookback window of N days so as to re-process those emails.

It might be necessary to manually delete the RawNotification records corresponding to the failed emails so that they get re-processed; another option would be to make the job smart enough to distinguish between "parsed successfully" and "parsing failed" notifications (by checking for the existence of a corresponding ParsedNotification record?) and re-processing the failed emails regardless?

jvanderaa:
## What's Changed

Adds two optional inputs to the **Update Circuit Maintenances** job so operators can re-fetch older notifications on demand, instead of always being limited to the normal incremental window (the timestamp of the last stored notification, or `raw_notification_initial_days_since` on the first run).

- **Days to look back** (`IntegerVar`) — fetch notifications from the last `N` days (`now - N days`).
- **Fetch notifications since** (`StringVar`, ISO 8601 date/time, e.g. `2026-01-31` or `2026-01-31T00:00:00Z`) — fetch notifications on or after that instant; assumed UTC when no timezone is given.

Either input overrides the normal watermark. When both are set, the **earliest** start time wins ("look back at least this far"). Leave both blank for normal incremental processing. An unparseable `Fetch notifications since` value fails the job with a clear error rather than silently doing nothing.

> **Note:** Nautobot's job framework has no `DateTimeVar` (vars are String/Text/Integer/Boolean/Choice/Object/File/IP/JSON), so the absolute datetime input is a `StringVar` parsed with `dateutil` and normalized to UTC.

### Re-processing already-stored notifications

A lookback window widens the range of notifications the job *fetches*, but `RawNotification` records are deduplicated by `(subject, provider, stamp)`, so an already-stored notification is skipped even inside the window. To **re-parse** notifications that were stored but failed to parse (e.g. after upgrading `circuit-maintenance-parser`), delete their `RawNotification` records first, then run the job with a lookback that covers them. This is documented in the user guide.

## Screenshots

Not applicable — this adds standard Nautobot-rendered inputs to an existing Job form (no custom UI).

## Testing

- Added tests for `get_since_reference` (days override, datetime override, both → earliest, first-run default) and for the job failing on an invalid `fetch_since` value; updated the existing watermark test for the `int` return type.
- `test_handler` (28) and `test_jobs` (4) pass; `ruff` lint + format clean; docs build (mkdocs strict) passes; no missing migrations.

## Docs

- `docs/user/app_use_cases.md` — new "Re-fetching older notifications (lookback window)" section documenting both inputs and the delete-then-reprocess workflow.

### DIFF
--- a/changes/170.added
+++ b/changes/170.added
@@ -1,0 +1,1 @@
+Added optional lookback window inputs (days to look back, fetch since) to the Handle Notifications job for re-fetching older notifications on demand.
--- a/docs/user/app_use_cases.md
+++ b/docs/user/app_use_cases.md
@@ -217,7 +217,21 @@ PLUGINS_CONFIG = {
 
 There is an asynchronous task defined as a **Nautobot Job**, **Handle Circuit Maintenance Notifications** that will connect to the email sources defined under the Notification Sources section (step above), and will fetch new notifications received since the last notification was fetched.
 
+By default the job runs **incrementally**: it only fetches notifications newer than the most recent `RawNotification` already stored. On the very first run (when no notifications exist yet) it looks back `raw_notification_initial_days_since` days instead (7 by default, see [the install guide](../admin/install.md)).
+
 Each notification will be parsed using the [circuit-maintenance-parser](https://github.com/networktocode/circuit-maintenance-parser) library, and if a valid parsing is executed, a new **Circuit Maintenance** will be created, or if it was already created, it will updated with the new data.
+
+#### Re-fetching older notifications (lookback window)
+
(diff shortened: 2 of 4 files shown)


glennmatthews (approved):
Looks reasonable to me.

PR #350, opened 2025-10-24
Refactored UIComponents Updated as per the previous comments

Pavan-183:
# Closes #<ISSUE NUMBER GOES HERE>
# What's Changed

1. Refactored **CircuitMaintenance**, **CircuitImpact**, **Note**, and **NotificationSource** model UI views to use the UI component framework.
2. Added **ParsedNotificationTable** to display ParsedNotification model data in tabular format, including Raw Subject, Provider, Source, Received timestamp, and JSON.

# Screenshots

| View                                        | Before                                                                                                                                                                                   | After                                                                                                                                                                                    |
| ------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| **List View (Circuit Maintenances)**        | image <br> image | image <br> image |
| **Detail View (Circuit Maintenances)**      | image <br>image | image <br> image <br>image |
| **Update View (Circuit Maintenances)**      | image <br> image | image <br> image |
| **Delete View (Circuit Maintenances)**      | image <br> image | image <br> image |
| **Bulk Edit View (Circuit Maintenances)**   | image <br> image | image <br> image |
| **Bulk Delete View (Circuit Maintenances)** | image <br> image | image <br> image |
| **Add View (Circuit Maintenances)**         | image <br> image | image <br> image |
| **Add View (Circuit Impact)**               | image <br> image | image <br> image |
| **Add View (Note)**                         | image <br> image | image <br> image |
| **List View (Notifications)**               | image <br> image | image <br> image |
| **Detail View (Notifications)**             | image | image |
| **List View (Notification Source)**         | image | image |
| **Detail View (Notification Source)**       | image | image |
| **Update View (Notification Source)**       | image <br> image | image <br> image |
| **Bulk Update View (Notification Source)**  | image | image <br> image |
| **Bulk Delete View (Notification Source)**  | image <br> image | image <br> image |
| **Delete View (Notification Source)**       | image | image |

# TODO

- [x] Explanation of Change(s)
- [ ] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/core/#creating-changelog-fragments))
- [ ] Attached Screenshots, Payload Example
- [ ] Unit, Integration Tests
- [ ] Documentation Updates (when adding/changing features)
- [ ] Example App Updates (when adding/changing features)
- [ ] Outline Remaining Work, Constraints from Design

PR #394, opened 2026-08-04
Address review follow-ups on the UIComponent refactor (#350)

jvanderaa:
## What this is

Follow-up review fixes on top of the merged #350 UI Component Framework refactor, targeting the `u/pavan-uicomponents-easy-kc` integration branch (not `develop`). #350 moved the detail views to the declarative framework and dropped the custom `*_retrieve.html` templates — good direction. This PR addresses one functional regression plus several cleanups our review flagged, and adds regression test coverage.

All changes are in `nautobot_circuit_maintenance/views.py` and `nautobot_circuit_maintenance/tests/test_views.py`.

### 1. Fix: `validate_source` no longer 500s on OAuth sources (the one blocker)
`Source.test_authentication()` deliberately re-raises `RedirectAuthorize` (`sources.py`) for OAuth sources (e.g. Gmail) that need to send the user through the provider consent flow. After the refactor, `validate_source` only caught `(AttributeError, TypeError, ValueError)`, so `RedirectAuthorize` escaped as an unhandled 500 — and the "Validate Authentication" button links straight to this action, so it was user-reachable.

`validate_source` now:
- Catches `RedirectAuthorize` and redirects to the OAuth authorize URL (restores pre-refactor behavior).
- Catches `ValueError` and **logs it** with `exc_info=True` (the refactor had dropped the logging).
- Attaches a Django `messages` flash (success/error) and **redirects back to the object detail view**, instead of `render()`-ing the legacy `notificationsource.html` template with three hand-built empty tables.

Why redirect-with-messages instead of the direct `render()`:
- It's the idiomatic pattern for a POST/GET action that reports a result, and it lets the detail page render via the UI framework rather than the legacy backward-compat template.
- It removes the "Temporary fix to pass test cases" scaffolding (the empty `ContactAssociationTable([])` / `DynamicGroupTable([])` / `ObjectMetadataTable([])`).
- Note: the previous implementation also supported a `return_url` GET param and redirected to it directly. I intentionally did **not** restore that — redirecting to a user-supplied URL is an open-redirect risk. We always redirect to the object's own detail page.

### 2. Cleanup: remove dead `CustomMaintenanceFieldsPanel`
This panel only existed to compute `data["notification_provider"]`, which is written but never read anywhere (no template or field references it). Replaced its single usage with a plain `ObjectFieldsPanel(fields="__all__")` and removed the now-unused `get_obj_from_context` import. No behavior change.

### 3. Cleanup: remove duplicated class attributes in `RawNotificationUIViewSet`
`serializer_class`, `table_class`, and `action_buttons` were each declared twice (once above `object_detail_content`, once below). Removed the redundant second block.

### 4. Cleanup + logging: `NotificationObjectFieldsPanel.get_data`
- Narrowed `except (AttributeError, TypeError, ValueError)` to `except ValueError` (what `Source.init()` actually raises) so unrelated bugs aren't masked, and added a `logger.warning(..., exc_info=True)` on the failure path.
- Dropped the defensive `try/except` around `instance.providers.all()` (a plain reverse manager access that doesn't raise those exceptions).
- Added a comment explaining that `account` / `source_type` / `providers_display` are runtime-only pseudo-fields attached to the instance for display.

### 5. Tests
- Updated `test_validate_view_ok` / `test_validate_view_ko` to assert the new redirect-and-flash behavior (302 to the detail view + the message in the messages framework) instead of a 200 render.
- Added `test_validate_view_redirect_authorize` — the regression test for the OAuth path: a source that raises `RedirectAuthorize` now redirects to `google_authorize` instead of erroring.
- Extracted the repeated permission-granting setup into a small `_grant_view_permission()` helper.

## Not included (flagging for discussion)
- No new changelog fragment: this all rolls up under the existing `changes/350.housekeeping` for the UIComponent effort, and the target is the integration branch, so a second fragment would just create a duplicate release-notes line.
- The legacy `notificationsource.html` template is now unused by the validate flow. I left it in place to keep this PR scoped to the regression + cleanups; we can remove it in the branch's eventual develop merge (it already carries a "3.0 TODO: remove" comment).

## Test evidence
- `invoke unittest --label nautobot_circuit_maintenance.tests.test_views` → `Ran 179 tests ... OK (skipped=29)`.
- `NotificationSourceTest` (verbose) → `Ran 24 tests ... OK (skipped=4)`, including `test_validate_view_ok`, `test_validate_view_ko`, and the new `test_validate_view_redirect_authorize`.
- `ruff check` / `ruff format --check` clean on both changed files.

### DIFF
--- a/nautobot_circuit_maintenance/tests/test_views.py
+++ b/nautobot_circuit_maintenance/tests/test_views.py
@@ -8,12 +8,14 @@ from unittest.mock import patch
 
 from django.conf import settings
 from django.contrib.contenttypes.models import ContentType
+from django.contrib.messages import get_messages
 from django.urls import reverse
 from nautobot.circuits.models import Circuit, CircuitType, Provider
 from nautobot.core.testing import ModelViewTestCase, ViewTestCases
 from nautobot.extras.models import Status
 from nautobot.users.models import ObjectPermission
 
+from nautobot_circuit_maintenance.handle_notifications.sources import RedirectAuthorize
 from nautobot_circuit_maintenance.models import (
     CircuitImpact,
     CircuitMaintenance,
@@ -311,16 +313,19 @@ class NotificationSourceTest(
         settings.PLUGINS_CONFIG = {"nautobot_circuit_maintenance": {"notification_sources": [cls.SOURCE_1.copy()]}}
         NotificationSource.objects.create(name=cls.SOURCE_1["name"])
 
-    @patch("nautobot_circuit_maintenance.handle_notifications.sources.IMAP.test_authentication")
-    def test_validate_view_ok(self, mock_test_authentication):
-        """Test for custom NotificationSourceValidate view."""
-        mock_test_authentication.return_value = True, "Test OK"
-
-        # Adding test to user to run Validate
+    def _grant_view_permission(self):
+        """Give the test user permission to view NotificationSource so the Validate action is reachable."""
         obj_perm = ObjectPermission(name="Test permission", actions=["view"])
         obj_perm.save()
         obj_perm.users.add(self.user)
         obj_perm.object_types.add(ContentType.objects.get_for_model(self.model))
+
+    @patch("nautobot_circuit_maintenance.handle_notifications.sources.IMAP.test_authentication")
+    def test_validate_view_ok(self, mock_test_authentication):
+        """A successful validation redirects to the detail view with a success message."""
+        mock_test_authentication.return_value = True, "Test OK"
+
+        self._grant_view_permission()
         source = NotificationSource.objects.get(name=self.SOURCE_1["name"])
         response = self.client.get(
             reverse(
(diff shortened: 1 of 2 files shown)


jvanderaa:
Superseded by #395, which re-homes this work onto `develop`.

The blocker here was not the changes in this PR — it was the base branch. `u/pavan-uicomponents-easy-kc` pins `nautobot = ">=2.4.16,<3.0.0"` and was cut before Nautobot 3.0 shipped, so its CI `stable` legs now fail at import with `ImportError: cannot import name 'PluginCustomValidator'` (renamed to `CustomValidator` in Nautobot 3) in `custom_validators.py` and `template_content.py` — files untouched by either this PR or #350. Since Nautobot 2 support lives on `ltm-2.4` and `develop`/`main` are Nautobot 3 only, there was no reason to keep the work on a 2.4-era base.

#395 carries all of these fixes plus the Nautobot 3 adaptations, and picked up one additional real bug in the process: the ParsedNotification `ObjectsTablePanel` badge link pointed at a nonexistent `parsednotification_list` route, which Nautobot 3 catches via `test_body_content_table_list_url`.

Closing in favor of #395.

PR #395, opened 2026-08-05
Refactor detail views to the UI Component Framework (#196)

### LINKED ISSUE #196: Switch to Object Detail View

jvanderaa:
Closes #196.

Re-homes the UI Component Framework refactor onto `develop`. The original work is @Pavan-183's (#350) and he is credited as co-author; this PR adapts it to Nautobot 3, incorporates review feedback, and adds test coverage.

## Why this targets `develop` rather than the integration branch

#350 was merged into the `u/pavan-uicomponents-easy-kc` integration branch. That branch pins `nautobot = ">=2.4.16,<3.0.0"` and was cut before Nautobot 3.0 shipped, so it can no longer be tested against a supported Nautobot — its CI `stable` legs fail at import with `ImportError: cannot import name 'PluginCustomValidator'` (renamed to `CustomValidator` in Nautobot 3), in files unrelated to the refactor.

Nautobot 2 support lives on `ltm-2.4`; `develop` and `main` are Nautobot 3 only. So there is no Nautobot 2 obligation for this change, and the stale imports simply disappear by building on `develop`, which already carries the migrated versions of those files.

The refactor itself is still needed — `develop` had not independently migrated. Its detail views still used `get_extra_context()` and all four `*_retrieve.html` templates were still present.

## What this does

- Moves the CircuitMaintenance, CircuitImpact, Note, RawNotification, and NotificationSource detail views to declarative `ObjectDetailContent` panels.
- Deletes the four custom `*_retrieve.html` templates and `note.html`, which the framework now renders. `parsednotification.html` is kept — `ParsedNotificationView` is still a plain `ObjectView`.
- Adds table columns and a `ParsedNotificationTable` for the new panels.

## Nautobot 3 adaptations (not in the original)

- **Imports come from the public `nautobot.apps.ui` surface** rather than `nautobot.core.ui.object_detail` / `nautobot.core.ui.choices`. All the components are exported there, and app development standards prefer `nautobot.apps.*`.
- **`enable_related_link=False` on the ParsedNotification panel.** `ParsedNotification` is detail-only — `urls.py` defines no list route — so the panel's badge link had no URL to resolve to and logged a warning on every Circuit Maintenance page view. Nautobot 3 enforces this via `test_body_content_table_list_url`, which failed until this was set. This is a genuine bug the 2.4-era branch could not surface.

## Review feedback incorporated from #350

- **`validate_source` no longer raises on OAuth sources.** `Source.test_authentication()` deliberately re-raises `RedirectAuthorize` for OAuth sources (e.g. Gmail) needing consent. The refactor caught only `(AttributeError, TypeError, ValueError)`, so it escaped as an unhandled 500 — and the "Validate Authentication" button links straight to that action. It now catches `RedirectAuthorize` and redirects to the provider consent flow, logs `ValueError` failures with `exc_info=True`, and reports the outcome via a `messages` flash plus a redirect to the detail view. This replaces the temporary scaffolding that rendered the legacy template with three hand-built empty tables.
  - The pre-refactor code also honored a `return_url` GET parameter and redirected to it. That is intentionally **not** restored — redirecting to a user-supplied URL is an open-redirect risk. The redirect always goes to the object's own detail page.
- Removed a dead `CustomMaintenanceFieldsPanel` whose computed `notification_provider` value was never read.
- Removed duplicated `serializer_class` / `table_class` / `action_buttons` declarations on `RawNotificationUIViewSet`.
- Narrowed over-broad exception handling in `NotificationObjectFieldsPanel.get_data` to `ValueError` and restored logging on the failure path.
- Addressed @gsnider2195's and @glennmatthews' notes from #350 (`ButtonActionColorChoices` for the Validate button, no out-of-scope view churn).

## Deliberately excluded

- Pavan's `forms.py` change swapped `CircuitMaintenanceBulkEditForm`'s bases to `(NautobotBulkEditForm, TagsBulkEditFormMixin)`. Mixins belong before the base form class, so `develop`'s existing order is kept.
- `develop` intentionally removed `custom_view_additional_permissions` from the `validate_source` action; that removal is preserved rather than reverted.
- Cosmetic template churn from the original branch (quote style, named `{% endblock %}`, stripped trailing newlines) is omitted to keep the diff reviewable.

## Testing

- Full suite: **294/296**, then `test_views` **196 OK (31 skipped)** after the `enable_related_link` fix.
- The one remaining failure is `test_basics.TestVersion` (`'3.0.1a0' != '3.1.2a0'`), which reproduces identically on a pristine `develop` checkout — a stale app version in the local container image, unrelated to this change.
- `ruff check` and `ruff format --check` clean.
- Adds `test_validate_view_redirect_authorize`, a regression test for the OAuth path, and updates the two existing validate tests for the redirect-and-flash behavior.

## Supersedes

- #394 (the follow-up fixes against the integration branch) — closing in favor of this.
- #350 remains merged on `u/pavan-uicomponents-easy-kc`; that branch can be retired once this lands.

## Screenshots

Detail-view rendering is unchanged in intent but now driven by panels rather than templates. Happy to attach before/after captures of the Circuit Maintenance, Notification Source, and Raw Notification detail pages if that would help review.


nautobot_circuit_maintenance/tables.py
```diff
@@ -63,11 +76,39 @@ class CircuitImpactTable(BaseTable):
     circuit = tables.Column(linkify=True)
     impact = tables.Column(linkify=True)
```
glennmatthews, line 77: Remove `maintenance` and `circuit` columns? They're no longer in `Meta.fields` - intentional?

nautobot_circuit_maintenance/tables.py
```diff
@@ -63,11 +76,39 @@ class CircuitImpactTable(BaseTable):
...
 
+    cid = tables.Column(accessor="circuit.cid", verbose_name="ID", linkify=True)
```
glennmatthews, line 79: IIRC the `.` pattern in accessors is deprecated by django-tables2, which prefers `__` instead.

nautobot_circuit_maintenance/tables.py
```diff
@@ -81,11 +122,25 @@ class Meta(BaseTable.Meta):
...
         model = Note
-        fields = ("pk", "maintenance", "title", "level", "comment", "last_updated")  # pylint:disable=nb-use-fields-all
+        fields = (
+            "pk",
+            "maintenance",
+            "level",
+            "title",
+            "last_updated",
+            "comment",
+        )  # pylint:disable=nb-use-fields-all
```
glennmatthews, line 132: unneeded comment?

nautobot_circuit_maintenance/tables.py
```diff
@@ -96,3 +151,28 @@ class Meta(BaseTable.Meta):
...
+class ParsedNotificationTable(BaseTable):
+    """Table to display ParsedNotification model."""
+
+    pk = ToggleColumn()
+    raw_notification_subject = tables.Column(
+        accessor="raw_notification.subject", verbose_name="Raw Subject", linkify=True
+    )
+    raw_notification_provider = tables.Column(accessor="raw_notification.provider", verbose_name="Provider")
+    raw_notification_source = tables.Column(accessor="raw_notification.source", verbose_name="Source")
+    stamp = tables.Column(accessor="raw_notification.stamp", verbose_name="Received")
```
glennmatthews, line 165: Missing from `Meta.fields`? Remove?

nautobot_circuit_maintenance/views.py
```diff
@@ -19,8 +27,9 @@
     ObjectView,
 )
 from nautobot.circuits.models import Circuit
+from nautobot.core.choices import ButtonActionColorChoices
```
glennmatthews, line 30: importable from `nautobot.apps.choices` (preferred)

nautobot_circuit_maintenance/views.py
```diff
@@ -19,8 +27,9 @@
...
+from nautobot.core.templatetags import helpers
```
glennmatthews, line 31: import individual template tags from `nautobot.apps.templatetags` instead.

nautobot_circuit_maintenance/views.py
```diff
@@ -267,6 +313,44 @@ class NoteUIViewSet(NautobotUIViewSet):
...
+        if key != "raw":
+            return super().render_value(key, value, context)
+        # Normalize to bytes
+        if isinstance(value, memoryview):
+            value = value.tobytes()
+        elif not isinstance(value, (bytes, bytearray)):
+            try:
+                value = bytes(value or b"")
+            except (TypeError, ValueError):
+                value = b""
```
glennmatthews, line 341: possible to add test coverage for this?

nautobot_circuit_maintenance/views.py
```diff
@@ -313,6 +417,52 @@ class ParsedNotificationView(ObjectView):
...
+
+    def get_data(self, context):
+        """Render providers as a list of hyperlinks."""
+        instance = context.get("object") or context.get(self.context_object_key)
+        # `account`, `source_type`, and `providers_display` are not model fields; they are attached to the
+        # instance here so the panel can surface source metadata that only exists at runtime via Source.init().
+        try:
+            source = Source.init(name=instance.name)
+            setattr(instance, "account", source.get_account_id())
+            setattr(instance, "source_type", source.__class__.__name__)
```
glennmatthews, line 431: Not obvious to me why we need to use `setattr()` here and below, versus just directly assigning `instance.account = source.get_account_id()`, etc.

nautobot_circuit_maintenance/views.py
```diff
@@ -313,6 +417,52 @@ class ParsedNotificationView(ObjectView):
...
+        """Add dynamic fields for account and providers."""
+        if key == "providers_display":
+            if not value:
+                return helpers.HTML_NONE
+            items = []
+            for val in value:
+                items.append(helpers.hyperlinked_object(val))
+            if not items:
+                return helpers.HTML_NONE
+            return format_html(
```
glennmatthews, line 459: Add test coverage? LInes 457-458 *look* unnecessary/unreachable to me.

